PR #5, opened 2016-09-12
Async spike

thardy:
I've altered the core functionality to use Promises.  I've also caused the HandleCache to call a new getTemplate method.  The default is to just do what was being done before, but this now allows consumers to override getTemplate to be whatever function they want.  It just needs to return a Promise.

I altered my changes to use any-promise, so consumers can use whatever promise library they want.

Note the it.only in test/tags.js.  This causes mocha to only run that particular test.  We'll need to remove that of course and alter all the other tags to work with the Promises.  I've only done include.

Let me know if you have any questions.  I can help in any way you need.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -9,41 +9,6 @@ var render = {
         assert(scope, 'unable to evalTemplates: scope undefined');
 
         var html = '';
-//        var promiseChain = Promise.resolve(''); // create an empty promise to begin the chain;
-//        templates.some((template, index) => {
-//            if (scope.get('forloop.skip')) return true;
-//            var promiseLink = Promise.resolve('');
-//            switch (template.type) {
-//                case 'tag':
-//                    // Add Promises to the chain that need to be resolved sequentially
-//                    promiseLink = this.renderTag(template, scope, this.register)
-//                        .then((partial) => {
-//                            if (partial === undefined) return true; // basically a noop (do nothing)
-//                            html += partial;
-//                        });
-//                    promiseChain = promiseChain.then(promiseLink); // add a link to the chain
-//                    break;
-//                case 'html':
-//                    promiseLink = Promise.resolve(template.value)
-//                        .then((partial) => {
-//                            html += partial;
-//                        });
-//                    promiseChain = promiseChain.then(promiseLink); // add a link to the chain
-//                    break;
-//                case 'output':
-//                    var val = this.evalOutput(template, scope);
-//                    promiseLink = Promise.resolve(val === undefined ? '' : stringify(val))
-//                        .then((partial) => {
-//                            html += partial;
-//                        });
-//                    promiseChain = promiseChain.then(promiseLink); // add a link to the chain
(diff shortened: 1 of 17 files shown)


src/render.js
```diff
@@ -1,30 +1,38 @@
...
 
     renderTemplates: function(templates, scope) {
         assert(scope, 'unable to evalTemplates: scope undefined');
-        var html = '',
-            partial;
-        templates.some(template => {
+        var htmlBlocks = [],
+            partial,
+            promises = [];
+        templates.some((template, index) => {
```
harttle: Since `render` returns Promises now, all tags require `renderTemplates` will need this tech to re-construct :)

harttle: Be careful not let scope variables evaluated asyncly, which makes the implementation of control flow like `for` and `if` trivial. Note that async fashion may import implementation issues that does not exist in sync fashion:

What if the client user `{%capture%}` a variable (which invokes `renderTemplate`) and use it as `{%if%}` condition? We may need the promises resolved in serial in this situation.

tags/for.js
```diff
@@ -57,14 +61,47 @@ module.exports = function(liquid) {
...
+                scopes.push(_.clone(scope));
                 scope.pop(ctx);
-
-                if (breakloop) return true;
             });
-            return html;
+
+            // This is some pretty tricksy javascript, at least to me.  Bluebird would have made this a lot easier, but
+            //  we are trying to not use anything that can't be done in native node Promises.
+            // This basically just processes an array of promises sequentially for every argument in the array - http://webcache.googleusercontent.com/search?q=cache:rNbMUn9TPtkJ:joost.vunderink.net/blog/2014/12/15/processing-an-array-of-promises-sequentially-in-node-js/+&cd=5&hl=en&ct=clnk&gl=us
```
harttle: you're right. We can extract these code and do polyfills to Promise later, to facilitate Promise upgrade.

test/tags.js
```diff
@@ -11,7 +11,11 @@ var liquid = Liquid({
     ctx, src, dst;
 
 function test(src, dst) {
-    expect(liquid.parseAndRender(src, ctx)).to.equal(dst);
+    liquid.parseAndRender(src, ctx)
```
harttle: You should return this promise here, and finally return this promise in `.it()` function. Otherwise, Mocha will assume test success for undefined return value without Exception. See: <https://mochajs.org/#working-with-promises>

thardy: Will do.  I haven't done any testing at all outside of one test at a time, and most of those are me actually debugging.  That's why I haven't looked into that yet.

I'm making good progress.  I pushed a commit to the asyncSpike branch of my fork just this morning that I'm hoping gets me past my biggest hurdle so far.  I'm ready to start testing everything again by removing the .only and getting the rest of the tags fixed up.  I expect to finish this weekend.

thardy:
Closing because this is superceded by another pull request.

PR #10, opened 2016-09-26
Conversion to Async

thardy:
All tests are passing except the express tests in tests/liquid.js.  I commented those out for now.


harttle:
Great work! Let me do the rest

PR #13, opened 2016-10-05
moved filepath manipulation into getTemplate

thardy:
This allows overridden versions of getTemplate to receive the filepath raw, and do whatever they want to with it.  For me, I'm just doing a lookup of the template name in Mongo.  I didn't want c:\whatever prepended and .liquid appended.  I needed to just see the raw template name as supplied in the include tag.


harttle:
merged and published at @1.2.3

PR #14, opened 2016-10-15
Interpolate bracket access

richgilbank:
There's probably a better way to do this, but it seems to work in a pinch. 
Previously, `getPropertyByPath` was replacing bracket access with `.`, then splitting. However, this doesn't allow it to resolve nested keys:
```
collections[settings.frontpage].products
```
was being treated as 
```
collections.settings.frontpage.products
```
Anyways, hope this helps 😄

### DIFF
--- a/src/lexical.js
+++ b/src/lexical.js
@@ -5,7 +5,7 @@ var quoteBalanced = new RegExp(`(?:${singleQuoted.source}|${doubleQuoted.source}
 
 var number = /(?:-?\d+\.?\d*|\.?\d+)/;
 var bool = /true|false/;
-var identifier = /[a-zA-Z_$][a-zA-Z_$0-9]*/;
+var identifier = /[a-zA-Z_\-$][a-zA-Z_$0-9\-]*/;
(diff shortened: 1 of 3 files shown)


richgilbank:
d'oh, clearly didn't run the tests locally - will try to take a look in a bit

harttle:
Should we allow hyphens in identifiers? I mean hyphens are generally not allowed in most of the programming languages, cause this disables the ability of subtract expressions. Or, does the official shopify liquid support this?

killalau:
The official liquid project in ruby support hyphens in variable name.
```
{% assign var-1 = 1 %}
{{ var-1 }} // return 1
```

I just made a small experiment and found that all this works...
```
{% assign var2- = 2 %}
{{ var2- }} // return 2

{% assign -var3 = 3 %}
{{ -var3 }} // return 3

{% assign 4-3 = 4 %}
{{ 4-3 }} // return 4

{% assign -5-5 = 5 %}
{{ -5-5 }} // return 5
```

BUT not number
```
{% assign -6 = 6 %}
{{ -6 }} // return -6
```

harttle:
Thank you for your experiments @killalau , I'll get this working this week.

PR #15, opened 2016-10-23
Template models

thardy:
Take a look at the changes I've made in order to use shopify-liquid with my CMS.  If these are too intrusive, let me know.  If you can think of a better way to modularize these changes and keep them separate from the base logic, I'll gladly do that.  The last option is - I can always just run off of a fork of your code, and these changes don't have to go back into your code at all.

Basically, everything works as normal, but if someone overrides getTemplate to return template objects instead of just string templates (e.g. `{ title: 'header', template: "<h1>{{title}}</h1>"}`).  TemplateObjects are just objects where all the properties are the model, and the template is expected to be inside a property named "template".  This allows me to easily save a partial with default values.  I then altered the include tag to unshift these partial models at the front of the scope stack.  That way, if a containing template doesn't have a title property, the property built into the partial itself will be used, but otherwise the containing template's properties will override the included partial's properties.

Once again, if this is too niche to merge into the base code, let me know, or if you can think of a more elegant way to extend the base with this functionality while keeping the base clean, let me know that too.

Thanks.


harttle:
As I understand it, you want to provide a default scope for each template string. I think this can be done by simply registering an `include` tag, in which you can push the default scope, render the template, and finally pop the default scope. (so I think you may not need the unshift operation?)

I mean if the push scope->render->pop scope logic can be done within the include tag, we should not modify the core code base.

The `include` tag can be maintained separately from the liquid-engine, this is recommended by design. For example, you can maintain a `filters/` and `tags/` directory in your CMS project, and register them with `.registerFilter()` and `.registerTag()` API. So in my opinion we need not to merge your customized `include` tag. 

If the push->render->pop sequence can not solve your problem, we add the `unshift` method to the scope object.

thardy:
The push and pop scenario doesn't work because I want the containing template's model to override the included template's model, hence the need for unshift.  With push, the included template's model will always win.

For this to work with multiple nested includes, the getTemplate in include.js also has to return templateObjects if the overridden getTemplate does.  That's the only way I've found to handle includes within includes and have the outer models override the inner model's properties.

I'm fine just using my fork.  If I come up with a more elegant way to extend the core code, I'll let you know.  Maybe we can just call an onTemplateReturned function that can be overridden by consumers?  I needed to replace what got cached as well though, so I would basically need to override the entire behavior found in include.js, which is what I did.  

I do think you'd want unshift and shift available on scope though for extensibility purposes.  I certainly had a need for it, and I can see other people wanting to write custom tags that might need it as well.

harttle:
I merged your scope.js and the corresponding test. Do you need the `templateObject` still?
I'm not quit understand your scenario, let's assume there's nested includes, say: 

`main.html` includes `foo.html`, while `foo.html` includes `bar.html`.

Before rendering `foo.html` in your include tag, unshift the default scope for `foo.html`.
Then comes to `bar.html`,
Before rendering `bar.html` in your include tag, you also unshift the scope for `bar.html`.

At this time the scope stack will be: `[main, foo, bar]`, so when rendering `bar.html` the scope stack is ready. Is this what you want for your template hierarchy?

thardy:
In the scenario you describe, after I unshift foo then unshift bar, the scope stack will be: [bar, foo, main].  That way main will win for any properties it has.  Any properties it does not have will come from foo next, then lastly, bar.

Here are the exact tests from my templateEngine class in my cms, which is using my fork of shopify-liquid under the covers...

```
    it("can render includes with their own model", function () {
        var templateString = "<p>{% include 'partialWithModel' %}</p>";
        var model = {};

        return expect(templateEngine.Render(templateString, model)).to.eventually.equal('<p>title is Default Title</p>');
    });

    it("can render includes with declared parameters", function () {
        var templateString = "<p>{% include 'partialWithModel', title: 'Hashed Title' %}</p>";
        var model = {};

        return expect(templateEngine.Render(templateString, model)).to.eventually.equal('<p>title is Hashed Title</p>');
    });

    it("can have outer template model override included template's model", function () {
        var templateString = "<p>{% include 'partialWithModel' %}</p>";
        var model = { title: 'Parent Title' };

        return expect(templateEngine.Render(templateString, model)).to.eventually.equal('<p>title is Parent Title</p>');
    });
```

```
The partialWithModel templateObject looks like this...
    {
        name: 'partialWithModel',
        title: 'Default Title',
        template: "title is {{title}}"
    }
```

This way, partials can have their own models, like an events widget that has the latest 5 events in its own model, but they can also have other elements that can be overridden by their containing template, like a header or just about anything.  The unshift is needed to guarantee that the outermost template models win.  Hopefully, I'll have a demo of my cms available for you very soon that will make it much more clear.

harttle:
Oh no, I mean [bar, foo, main] actually. Then you want properties in `foo` override the ones in `bar` only if the property with same name is defined in `bar`? Does it bother when all properties come to `bar`?

Generally, the hash syntax is used to override properties within the included template. It seems that this is not enough for you. Looking forward to your demo.

harttle:
Since it'll be more clean for this to be done in a custom Tag or forked project, closing.

PR #21, opened 2017-03-15
reversed position

chenos:
eg:

```twig
{% for i in (1..15) reversed limit:4 offset:2 %}
  {{ i }}
{% endfor %}

{% for i in (1..15) limit:4 reversed offset:2 %}
  {{ i }}
{% endfor %}

{% for i in (1..15) limit:4 offset:2 reversed %}
  {{ i }}
{% endfor %}
```


harttle:
Good catch! Published @1.7.2

PR #25, opened 2017-03-19
Size can also be used with dot notation

chenos:
Size can also be used with dot notation (for example, {{ my_string.size }})

### DIFF
--- a/src/scope.js
+++ b/src/scope.js
@@ -76,17 +76,16 @@ var Scope = {
             throw new TypeError('undefined variable');
         }
         var variable = obj[varName];
-        var len = paths.length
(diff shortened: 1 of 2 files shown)


harttle:
In my understanding, it would be more clear to use `.length` to read the `length` property, rather than to read a `.size` property and do special calculations.

This will be rather confusing when there's already a `size` property exists there.

chenos:
`.size` is designed for cross platform compatibility, ensure that syntax is the same as native liquid(ruby), javascript environment `.size` is `.length` alias.

https://shopify.github.io/liquid/filters/size

harttle:
OK, sounds good. Published @1.7.5

PR #26, opened 2017-03-20
All values in Liquid are truthy except nil and false.

chenos:
For details, see https://shopify.github.io/liquid/basics/truthy-and-falsy/

### DIFF
--- a/src/syntax.js
+++ b/src/syntax.js
@@ -47,7 +47,7 @@ function isTruthy(val) {
 }
 
 function isFalsy(val) {
-    return false === val || undefined === val;
+    return false === val || undefined === val || null === val;;
 }
 
 module.exports = {
--- a/src/syntax.js
+++ b/src/syntax.js
@@ -43,12 +43,11 @@ function evalValue(str, scope) {
 }
 
 function isTruthy(val) {
-    if (val instanceof Array) return !!val.length;
-    return !!val;
+    return !isFalsy(val);
 }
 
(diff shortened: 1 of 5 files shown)


src/syntax.js
```diff
@@ -43,12 +43,11 @@ function evalValue(str, scope) {
...
 
 function isTruthy(val) {
-    if (val instanceof Array) return !!val.length;
-    return !!val;
+    return !isFalsy(val);
 }
 
 function isFalsy(val) {
-    return !isTruthy(val);
+    return false === val || undefined === val;
```
harttle: what about `null` and `0`?

test/tags/for.js
```diff
@@ -80,13 +87,13 @@ describe('tags/for', function() {
...
 
-    it('should support for reversed in the middle position', function() {
+    it('should support for reversed in the first position', function() {
         var src = '{% for i in (1..5) reversed limit:2 %}{{ i }}{% endfor %}';
         return expect(liquid.parseAndRender(src, ctx))
             .to.eventually.equal('21');
     });
 
-    it('should support for reversed in the first position', function() {
+    it('should support for reversed in the middle position', function() {
```
harttle: good catch

src/syntax.js
chenos: In my understanding, ruby's `nil` is `null` or `undefined` in javascript.

So

```js
return false === val || undefined === val || null === val;
```

PR #27, opened 2017-03-21
Comparison on NULL is always false

chenos:
https://github.com/Shopify/liquid/blob/master/test/integration/tags/if_else_tag_test.rb#L114


harttle:
published @1.7.6

PR #28, opened 2017-04-10
support multiline tags

wojtask9:
It would be great that shopify-liquid could handle multiline tags for example

{% tag d1:'a'
           d2:'b'
           d3:'c'
%}

This is useful if we have very long lines of parameters/attributes passed to tag

In JS there isn't "dotall" modifier but this can be solved using [\s\S] modifier to replace ".".

See more http://stackoverflow.com/questions/1068280/javascript-regex-multiline-flag-doesnt-work


harttle:
Published @1.7.7

PR #31, opened 2017-07-12
Remove unused `stop` and `skip` from `forloop` object

aleclarson:
The `stop` and `skip` properties of `forloop` are unused since this commit: 951225fb75141a235b4d2e52de4b8dda585a1ccd


harttle:
Good catch, thank you!

PR #32, opened 2017-07-12
Iterate keys of object when used with `for...in` loop

aleclarson:
This allows `for...in` loops to work with `{}` or `Object.create(null)`.

I added `isObject` and `isEmpty` for cleaner `if` statements. I can remove those if you prefer.

### DIFF
--- a/tags/for.js
+++ b/tags/for.js
@@ -1,6 +1,7 @@
 const Liquid = require('..')
 const lexical = Liquid.lexical
 const mapSeries = require('../src/util/promise.js').mapSeries
+const isPlainObject = require('../src/util/underscore.js').isPlainObject
 const RenderBreakError = Liquid.Types.RenderBreakError
 const assert = require('../src/util/assert.js')
 const re = new RegExp(`^(${lexical.identifier.source})\\s+in\\s+` +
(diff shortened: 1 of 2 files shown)


tags/for.js
```diff
@@ -38,8 +38,10 @@ module.exports = function (liquid) {
     render: function (scope, hash) {
       var collection = Liquid.evalExp(this.collection, scope)
 
-      if (!Array.isArray(collection) ||
-                (Array.isArray(collection) && collection.length === 0)) {
+      if (isObject(collection)) {
```
harttle: Can you move these language related functions to [/src/util/underscore.js][underscore-local].

I don't mind renaming `underscore.js` to `lang.js` or something more appropriate if you find this file name confusing, cause it's actually not the [underscore][underscore] libarary.

[underscore-local]: https://github.com/harttle/shopify-liquid/blob/master/src/util/underscore.js
[underscore]: http://underscorejs.org/

tags/for.js
```diff
@@ -90,3 +92,18 @@ module.exports = function (liquid) {
...
+
+function isObject(collection) {
+  if (collection) {
+    const ctr = collection.constructor
+    return ctr === Object || ctr == null
+  }
+  return false
+}
+
+function isEmpty(collection) {
```
harttle: `isEmpty` should also capable of none-empty objects.

The underscore/lodash can be implementation references, e.g. <https://lodash.com/docs/4.17.4#isEmpty>

aleclarson:
I added `isPlainObject` to `src/util/underscore.js` and removed `isEmpty` in favor of just simplifying the `if` statement.

`{}` and `Object.create(null)` with no keys are accounted for, since they are converted to arrays.

harttle:
I've merged your commits and added test cases. Behaviors related to empty and non-empty strings will be compliant with [the ruby version](https://github.com/Shopify/liquid).

This feature is released as 1.9.0. Thank you to lead this feature.

PR #34, opened 2017-07-20
add identation value to token

wojtask9:
Adds indentation to token.
This information could be useful for custom tags.

### DIFF
--- a/src/tokenizer.js
+++ b/src/tokenizer.js
@@ -37,7 +37,7 @@ function parse (html, filepath, options) {
       token.args = match[2]
 
       // get last line indentation
-      let lineStart = (htmlFragment || '').split('\n')
+      var lineStart = (htmlFragment || '').split('\n')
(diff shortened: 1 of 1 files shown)


harttle:
It seems that token.indent will be `3` for this case:

```liquid
foo{%if%}
```

Is this behavior correct or useful?

wojtask9:
this behavior is correct.
It is useful if you writing custom components.
Then in render you can access to indent value for current tag and add whitespaces at beginning to output
For example:
```
<div>
   {%cinclude%}
   <span>aaa</span>
   <div>bbb</div>
   {%endcinclude%}
</div>
```
if you use indent information and add whitespaces at beginning output will be:
```
<div>

   <span>aaa</span>
   <div>bbb</div>

</div>
```
Whitespaces at beginning are preserved

harttle:
Will be published in patch, thanks for your contribution

PR #35, opened 2017-08-02
Bug fix for `foo.bar[0].foo` that is not working.

jaswrks:
Corrects a bug where `foo.bar[0].foo` results in error: 'Cannot read property `foo` of undefined', because the sequence contains an empty string; i.e., `seq.push('')` should be avoided.

Another possible solution would be to strip empty strings from the final `seq` variable, but this PR fixes the immediate problem.


harttle:
Thank you for this fix, published @1.9.2

PR #43, opened 2017-10-18
Fixes #42: make `for` tag iterate objects like Shopify liquid.

### LINKED ISSUE #42: Iterating a hash
Hello

I have been trying to iterate a hash/object in Liquid but the output is not as expected.

See Liquid reference: https://github.com/Shopify/liquid/wiki/Liquid-for-Designers#allowed-collection-types

The reference says:

> When iterating a hash, item[0] contains the key, and item[1] contains the value:

```liquid
{% for item in hash %}
  {{ item[0] }}: {{ item[1] }}
{% endfor %}
```

However, `item[0]` and `item[1]` contain the first and second characters of the item key instead of having the key and the value of current item.

Am I iterating the hash correctly? Is this a not-implemented feature?

PS: Thank you for the work on liquidjs. Its awesome. 😃

thehappybug:
Ref: https://github.com/Shopify/liquid/wiki/Liquid-for-Designers#allowed-collection-types

### DIFF
--- a/tags/for.js
+++ b/tags/for.js
@@ -43,7 +43,7 @@ module.exports = function (liquid) {
         if (_.isString(collection) && collection.length > 0) {
           collection = [collection]
         } else if (_.isObject(collection)) {
-          collection = Object.keys(collection)
+          collection = Object.keys(collection).map((key) => [key, collection[key]])
         }
       }
       if (!Array.isArray(collection) || !collection.length) {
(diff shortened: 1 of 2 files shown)


thehappybug:
@harttle Is it possible to merge this?

harttle:
Good work. Thank you for reporting and fixing this bug. Will be published in next version.

harttle:
Published @1.9.7

PR #52, opened 2017-12-20
Ignore "fs" module in browser environments

strax:
Depending on this package causes an error in Webpack and other similar browser-side module bundlers as the Node.js core modules are not available. This change simply tells the bundler to ignore the fs package in browser environments.


harttle:
Thank you for this fix.

PR #70, opened 2018-05-08
Ensure strip_html filter uses same regex as Ruby Liquid. Fixes #68

### LINKED ISSUE #68: strip_html filter does not remove elements that include attributes
The `strip_html` filter is not comparable to the same filter in the Ruby version. While Ruby Liquid removes all HTML elements, this implementation currently ignores any opening tag containing attributes.

Comparing the regex, we can see why. [Ruby Liquid](https://github.com/Shopify/liquid/blob/0c802aba175aefb5cfb7e40c13c892075e70ffa5/lib/liquid/standardfilters.rb#L104):

```ruby
def strip_html(input)
    empty = ''.freeze
    input.to_s.gsub(/<script.*?<\/script>/m, empty).gsub(/<!--.*?-->/m, empty).gsub(/<style.*?<\/style>/m, empty).gsub(/<.*?>/m, empty)
end
```

As well as covering `script`, `style` and comment tags, it also removes any string of characters starting with `<` and ending with `>`.

[Liquid JS](https://github.com/harttle/liquidjs/blob/master/filters.js#L69) on the other hand uses a different approach:

```js
'strip_html': v => stringify(v).replace(/<\/?\s*\w+\s*\/?>/g, ''),
```

My JavaScript skills are pretty rough, but I monkey patched this filter to use the following regex instead:

```js
str.replace(/<script.*?<\/script>|<!--.*?-->|<style.*?<\/style>|<.*?>/g, '')
```

This more closely matches that used in the Ruby implementation. Would a PR with this change be accepted at all?

Thanks,

Paul

paulrobertlloyd:
See the discussion in #68. I hope my tests are correct; I used `&quot;` for my attribute test given [the advice I found here](https://stackoverflow.com/a/2004178).


aciccarello:
Should this handle tags on different lines?

The regex removes this:

```html
<style type="text/css">.test-one-line {display: none;}</style>
```

But not this
```html
<style type="text/css">
.test-one-line {display: none;}
</style>
```

harttle:
@aciccarello please try v10.6.1

aciccarello:
I've confirmed this is working as expected. Thanks!

PR #78, opened 2018-07-17
fix: correctly understand variables with question marks

### LINKED ISSUE #77: Variable with question mark does not parse
In Ruby, a question mark is a valid identifier character and is commonly used to denote a boolean. However, LiquidJS throws an error.

For example, this syntax is valid Liquid:

```
{% if form.posted_successfully? %}
  <div class="success">Success!</div>
{% endif %}
```

But throws an error:

```
RenderError: cannot eval 'form.posted_successfully?' as value, file:/home/ryan/Projects/.../contact-form.liquid, line: 27
  at renderTemplate.call.then.catch.e (/home/ryan/Projects/.../node_modules/liquidjs/src/render.js:21:17)
  at <anonymous>:null:null

From TypeError: cannot eval 'form.posted_successfully?' as value
  at evalValue (/home/ryan/Projects/.../node_modules/liquidjs/src/syntax.js:43:9)
  at Function.evalExp (/home/ryan/Projects/.../node_modules/liquidjs/src/syntax.js:30:10)
  at Object.render (/home/ryan/Projects/.../node_modules/liquidjs/tags/if.js:35:27)
  at Promise.resolve.then (/home/ryan/Projects/.../node_modules/liquidjs/src/tag.js:27:48)
  at <anonymous>:null:null
```

ryaninvents:
Fixes #77

### DIFF
--- a/src/lexical.js
+++ b/src/lexical.js
@@ -10,7 +10,7 @@ var number = /-?\d+\.?\d*|\.?\d+/
 var bool = /true|false/
 
 // peoperty access
-var identifier = /[\w-]+/
+var identifier = /[\w-]+[?]?/
 var subscript = new RegExp(`\\[(?:${quoted.source}|[\\w-\\.]+)\\]`)
 var literal = new RegExp(`(?:${quoted.source}|${bool.source}|${number.source})`)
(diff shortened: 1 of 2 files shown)


ryaninvents:
Thanks @harttle!

PR #88, opened 2018-08-24
Fix `name` variable type change

loopal:
`name` should be a string. However, `this.get(name)` will return an integer, which is later assigned to `name`. Hereby, it will fail at `if (name.length)` in `push ()`.

### DIFF
--- a/src/scope.js
+++ b/src/scope.js
@@ -110,7 +110,7 @@ var Scope = {
             assert(j !== -1, `unbalanced []: ${str}`)
             name = str.slice(i + 1, j)
             if (!lexical.isInteger(name)) { // foo[bar] vs. foo[1]
-              name = this.get(name).toString()
+              name = (this.get(name) ? this.get(name) : '').toString()
             }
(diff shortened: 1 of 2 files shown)


src/scope.js
```diff
@@ -110,7 +110,7 @@ var Scope = {
             assert(j !== -1, `unbalanced []: ${str}`)
             name = str.slice(i + 1, j)
             if (!lexical.isInteger(name)) { // foo[bar] vs. foo[1]
-              name = this.get(name)
```
harttle: What if `this.get(name)` returns `undefined`? can you provide more test cases here

loopal: If `this.get(name)` returns `undefined`, `name = this.get(name).toString()` will throw an error. 
We can change `name = this.get(name)` to `name = (this.get(name) ? this.get(name) : '').toString()`. 
Thus, if `this.get(name)` returns `undefined`, `name` will be assigned `''`, and no error would be thrown.

PR #96, opened 2018-11-21
fix assign tag regex to include linebreaks in value

thelornenelson:
This fixes a bug where any filters beyond the first line of an assign tag were ignored, due to the differences between the `.` regex character in Ruby and JS. In Ruby it matches any character, but in JS it matches any character **other than a newline**.


harttle:
Thank you for your contribution! Published liquidjs@6.1.1

Issue #65, opened 2018-03-20
Promise support in scope.get


cmawhorter:
The ruby implementation of liquid supports async/lazy data access so that data can be loaded on-demand.

This is described here in the rust version of of liquid: https://github.com/cobalt-org/liquid-rust/issues/118#issuecomment-304545093

`scope.get` should work with promises so that I could do {{ some_async_product.price }} and have the product data load via API call that returns a promise.  

This would be a big, breaking change, but I've implemented it in [our fork](https://github.com/stampr/liquidjs/blob/125cbdd33a6decedfe471c7aa0397246fc7b66c2/src/scope.js#L43) and it was definitely not easy, but at this point it is working well.  (I do worry about race conditions with scope.push/pop, though it hasn't been a problem yet.)

harttle:
If `getFromContext` is called only when the corresponding block is rendering, race condition can be avoided since, typically, `scope.push` is called before the block rendering, and `scope.pop` is called after the block has been rendered.
It's great to have promise support in scope.get. We can open a branch to do that.

cmawhorter:
FYI - i plan to refactor into async/await once IE Edge ships chromium.  that might be a good time to figure out how to merge, if you're interested.

harttle:
That would be great! We've shipped to ES6 and introduced babelify since [v5.3.0-0](https://github.com/harttle/liquidjs/commit/4ece4e208f49e56267e01f6a5a5ac6b53c84ad11) to generate a fully compatible ES5 version for #83 . I can help if there're any conflict with the recent features.

harttle:
Since it'll introduce many micro-tasks (maybe slower) to support this feature, I' d rather make it pending until we got benchmark ready. I opened a issue #105 to get it rolling.

harttle:
It's supported on 8.0.0. Refer to [this test case](https://github.com/harttle/liquidjs/blob/84ea2c74882c4b3b59d2325b1da0f7821565de24/test/integration/drop/drop.ts#L44-L55)

Issue #48, opened 2017-11-29
Offer a renderSync method?


zachleat:
Most templating engines offer synchronous render methods (ejs, mustache, handlebars, haml, pug, nunjucks, etc).

Does one exist for liquidjs? If not, could one be added?

(love this library btw—thank you)

harttle:
Maybe not possible current code base. We know async interface cannot be converted to sync. So we made intensive discussion over whether to support async in #3 .

Nevertheless, supporting sync render methods without regression would be great. Welcome to propose a solution for this issue.

zachleat:
Interesting! To be frank I think I’d be happy with throwing an error if the template has an async tag and rendering is in sync mode.

zachleat:
I worked around this in my upstream code so it isn’t necessary for me any more! It may be useful for someone else in the future but I would consider it low priority. In fact, other libs might tend towards what you’ve done instead 😎 Thanks for the discussion!

harttle:
Now we use TypeScript and I'll try make it support both async and sync calls, maybe generate additional code during rollup build.

harttle:
`renderFileSync`, `renderSync`, `parseAndRenderSync` are now supported by v9.0.0, but with some breaking changes. Feel free to file an issue if there's problem during upgrade. Closing, hope it helps!

Munawwar:
Ok I did a simple test of `renderSync()` and seems to be working. Integrated liquidjs with SahteReact (front-end library I was building for Shopify stores... so now, one can reuse shopify templates on the client-side co-existing with libraries like jquery!)

https://github.com/Munawwar/SahteReact/commit/3251a213553e1ce719a0a36f3cd97e91cfa745a5

@harttle how to `include` partials on browser (without async fetching)? how to use `dynamicPartials`?

harttle:
By default, it's implemented as sync xhr  which is absolutely bad practice. see:

https://github.com/harttle/liquidjs/blob/f661cd8b858fa897e595978bcb1eafb1137ee357/src/fs/browser.ts#L47 , 

A customed, platform-related `fs` can be specified in the [LiquidOptions](https://harttle.land/liquidjs/interfaces/_liquid_options_.liquidoptions.html#fs), as discussed in #131

PR #111, opened 2019-03-01
Add sync mode

burashka:
this.parse is sync method. This API doesn't need async. Please let me use it in sync.

### DIFF
--- a/src/liquid.ts
+++ b/src/liquid.ts
@@ -41,6 +41,10 @@ export default class Liquid {
     const scope = new Scope(ctx, options)
     return this.renderer.renderTemplates(tpl, scope)
   }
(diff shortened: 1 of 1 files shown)


harttle:
The `render()` is async and returns a promise, thus `parseAndRender()` will return a promise. Actually, it's intended to be async to embrace the JavaScript way, see: #3 and #5 .

By the way, we are planning to make scopes to be async too, see #65.

burashka:
Ok, sorry

harttle:
Nothing to sorry about :)  There's once a discussion over `renderSync()` method (#48). It could be possible to provide a sync version and throw an Error once async tags are called. It's a good point.

PR #114, opened 2019-03-20
Fix Register Tags example

ssendev:
the example didn't work. this fixed it for me

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,7 +228,9 @@ engine.registerTag('upper', {
     },
     render: function(scope, hash) {
         var str = Liquid.evalValue(this.str, scope); // 'alice'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -228,7 +228,9 @@ engine.registerTag('upper', {
     },
     render: function(scope, hash) {
         var str = Liquid.evalValue(this.str, scope); // 'alice'
-        return Promise.resolve(str.toUpperCase()); // 'Alice'
+        return str.then(function(name) {
```
harttle: Thank you for your work! Maybe better to use async function and await that value.

ssendev:
Well in my code i used async ;) but the example used Promise so i kept it that way. Anyway, here is an updated  version.

PR #139, opened 2019-07-08
Add handle filter

ButsAndCats:
Shopify specific string filter https://help.shopify.com/en/themes/liquid/filters/string-filters#handle-handleize


harttle:
Thank you for your comtribution, but `handleize` is not a [standard filter](https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb). Maybe better moved into a separated repository and import via the [plugin API](https://github.com/harttle/liquidjs#plugin-api) into liquidjs. Feel free to add your plugin into that list when you're ready.
I'm sorry for that but I'll make it clear in the contribution guidelines.

PR #151, opened 2019-08-08
Fix quotation tokenizing

### LINKED ISSUE #148: Object property access is not parsed correctly
When trying to access object's property with a name containing special character, the parsing fails. See for example
```js
const raw = {{ yarr["foo:zoo"] }};
const params = {yarr: {"foo:zoo": "Bar"}};
engine
	.parseAndRender(raw, params)
	.then(console.log)
```

AST is something like 
```js
Output {
        token:
         OutputToken {
           trimLeft: false,
           trimRight: false,
           type: 'output',
           col: 1,
           line: 1,
           raw: '{{ yarr["foo:zoo"] }}',
           value: 'yarr["foo:zoo"]',
           input: '{{ yarr["foo:zoo"] }}',
           file: undefined },
        value:
         Value { filters: [], strictFilters: false, initial: 'yarr["foo' } }
```

The `initial` was not parsed properly. This used to work in liquidjs 6.x.x

wyozi:
Fixes #148

### DIFF
--- a/src/template/value.ts
+++ b/src/template/value.ts
@@ -69,7 +69,12 @@ export default class Value {
         tokens.push(str[i++])
       } else {
         const j = i++
-        for (; i < str.length && !/[|,:\s]/.test(str[i]); ++i);
+        let ch
+        for (; i < str.length && !/[|,:\s]/.test(ch = str[i]); ++i) {
+          if (ch === '"' || ch === "'") {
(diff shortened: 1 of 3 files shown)


wyozi:
@harttle could you release a new version?

harttle:
@all-contributors please add @wyozi  for code

PR #154, opened 2019-08-24
fix: escape filter when input is undefined

jamelait:
The built-in escape filters renders "undefined" when the input is undefined :
````javascript
// myObj does not exist
const ret = await engine.parseAndRender(`{{ myObj.someProp | escape }}`)
expect(ret).toBe("") //  Expected: "" but Received: "undefined"
````
This PR fixes it.

### DIFF
--- a/src/builtin/filters/html.ts
+++ b/src/builtin/filters/html.ts
@@ -1,3 +1,5 @@
+import { stringify } from '../../util/underscore'
+
 const escapeMap = {
   '&': '&amp;',
   '<': '&lt;',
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/builtin/filters/html.ts [hunk 1]
```diff
@@ -14,7 +14,7 @@ const unescapeMap = {
 }
 
 function escape (str: string) {
-  return String(str).replace(/&|<|>|"|'/g, m => escapeMap[m])
+  return str ? String(str).replace(/&|<|>|"|'/g, m => escapeMap[m]) : ''
```
harttle: Try use [stringify](https://github.com/harttle/liquidjs/blob/master/src/util/underscore.ts) instead.

harttle:
@all-contributors please add jamelait for code

PR #171, opened 2019-11-10
Add ability to pass JSON context to CLI

brandonpittman:
If you pass a JSON file or JSON string as the first argument to the CLI, you can use variables in your templates that get parsed by the CLI.


harttle:
@all-contributors please add @brandonpittman for code.

PR #182, opened 2020-01-08
perf: prevent multiple case evaluations

mastodon0:
The case condition was evaluated for every branch


harttle:
@all-contributors please add mastodon0 for code

PR #189, opened 2020-02-14
Add `funding` entry to show up in `npm fund`

zachleat:
For example liquidjs is missing when I run:
image


harttle:
Thank you for your PR. Since liquidjs is auto-published by semantic-release, we need commit messages to be consistent.

Please re-submit with a format like `fix: add funding entry...`, it's checked on travis by commitlink, see: https://github.com/conventional-changelog/commitlint/#what-is-commitlint

PR #191, opened 2020-02-15
fix: keep orignal order if inputs are nil for `caseInsensitiveCmp`

richardo2016:
in function `caseInsensitiveCmp`, to keep original order when other items with specified property `undefined`, we should use logic like below:

```typescript
function caseInsensitiveCmp (a: any, b: any) {
    if (!a && !b) return 1
    else if (!a) return 1
    else if (!b) return -1
    ...
}
```

simplify it, we got
```typescript
function caseInsensitiveCmp (a: any, b: any) {
    if (!a) return 1
    if (!b) return -1
    ...
}
```

### DIFF
--- a/src/builtin/filters/math.ts
+++ b/src/builtin/filters/math.ts
@@ -19,8 +19,8 @@ export default {
 }
 
 function caseInsensitiveCmp (a: any, b: any) {
-  if (!b) return -1
   if (!a) return 1
+  if (!b) return -1
   a = toLowerCase.call(a)
   b = toLowerCase.call(b)
   if (a < b) return -1
--- a/test/integration/builtin/filters/math.ts
+++ b/test/integration/builtin/filters/math.ts
@@ -99,6 +99,16 @@ describe('filters/math', function () {
       const html = await l.parseAndRender(src, { students })
       expect(html).to.equal('alice bob carol')
     })
+    it('keep orignal order based on specified property', async () => {
+      const src = '{{ students | sort_natural: "age" | map: "name" | join }}'
+      const students = [
+        { name: 'bob' },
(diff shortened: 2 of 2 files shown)


harttle:
Your newly added case failed, see travis log. I tried shopify/liquid but find it rather confusing:

```
// scope in ruby
scope = { "students" => [
    { "name" => 'bob' },
    { "name" => 'alice', "age" => 2 },
    { "name" => 'amber' }
]}

// template
{{ students | sort_natural: "age" | map: "name" | join }}

// output
alice amber bob
```

It seems that those items without the specified property are moved to the tail, and they are re-ordered alphabetically. I'm not sure whether keys in Ruby Map are ordered or not, but certainly it's not behaving the same way as JavaScript.

harttle:
Finally I find out that in Ruby all items with a `null` property value goes to the tail of the sorted list. I altered the test case a bit and `return 0` for `a == null && b == null` to keep the original order. There're still differences in sorting and property iteration order, I added some notes in [README.md](https://github.com/harttle/liquidjs#differences-and-limitations).

Though this PR can't be merged due to the test case not passing, thank you for pointing this out and initiating the test case, @all-contributors please add richardo2016 for code

richardo2016:
> Finally I find out that in Ruby all items with a `null` property value goes to the tail of the sorted list. I altered the test case a bit and `return 0` for `a == null && b == null` to keep the original order. There're still differences in sorting and property iteration order, I added some notes in [README.md](https://github.com/harttle/liquidjs#differences-and-limitations).
> 
> Though this PR can't be merged due to the test case not passing, thank you for pointing this out and initiating the test case, @all-contributors please add richardo2016 for code

thx for your attention to this feature!

PR #206, opened 2020-03-23
fix: return variable name in include error

CriGoT:
Adds a `getText()` missing in #205. Without it the error message started always with `illegal filename "[object Object]"`


harttle:
@all-contributors please add CriGoT for code.

PR #209, opened 2020-03-27
Add date term calculation

sumanth-spotdraft:
Allow date and term calculation in liquid

(A) If date = 31 March 2020 then 30 day period = Print date + 30 days . 
(B) If Date 1 = 31 March 2020 and Date 2 = 30 March 2021 then Print Duration between Date 1 and Date 2


harttle:
Thank you for your work!

If there're multiple features to be added, could you please submit different PRs so we can tackle them one by one. Due to recent performance updates (see #202, #205), there can be many conflicts. Additionally, here's a contribution guideline that may help: https://github.com/harttle/liquidjs#contribute-guideline

PR #213, opened 2020-04-02
fix: respect cache render options

mastodon0:
I tried to disable the cache for a single renderFile call and discovered this bug. 
The passed in cache option would always be ignored.

### DIFF
--- a/src/liquid-options.ts
+++ b/src/liquid-options.ts
@@ -89,11 +89,13 @@ export function normalize (options?: LiquidOptions): NormalizedOptions {
   if (options.hasOwnProperty('root')) {
     options.root = normalizeStringArray(options.root)
   }
-  let cache: Cache<Template[]> | undefined
-  if (typeof options.cache === 'number') cache = options.cache > 0 ? new LRU(options.cache) : undefined
-  else if (typeof options.cache === 'object') cache = options.cache
-  else cache = options.cache ? new LRU<Template[]>(1024) : undefined
-  options.cache = cache
+  if (options.hasOwnProperty('cache')) {
+    let cache: Cache<Template[]> | undefined
+    if (typeof options.cache === 'number') cache = options.cache > 0 ? new LRU(options.cache) : undefined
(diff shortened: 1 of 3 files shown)


src/liquid-options.ts
```diff
@@ -89,11 +89,13 @@ export function normalize (options?: LiquidOptions): NormalizedOptions {
...
-  else if (typeof options.cache === 'object') cache = options.cache
-  else cache = options.cache ? new LRU<Template[]>(1024) : undefined
-  options.cache = cache
+  if (options.hasOwnProperty('cache')) {
+    let cache: Cache<Template[]> | undefined
+    if (typeof options.cache === 'number') cache = options.cache > 0 ? new LRU(options.cache) : undefined
+    else if (typeof options.cache === 'object') cache = options.cache
+    else cache = options.cache ? new LRU<Template[]>(1024) : undefined
+    options.cache = cache
+  }
```
harttle, line 98: Is this change necessary?

test/integration/liquid/cache.ts
```diff
@@ -175,4 +190,19 @@ describe('LiquidOptions#cache', function () {
       expect(y).to.equal('foo')
     })
   })
+  it('should respect passed in cache=false option', async function () {
```
harttle, line 193: seems in the wrong level, should be inside describe renderFileSync

src/liquid-options.ts
mastodon0: Yes, the change in [liquid.ts#78](https://github.com/harttle/liquidjs/blob/90b863104e721a6041cf9d0f75de99d0d2767dda/src/liquid.ts#L78) would make couple of other tests crash. The options-object would always have the cache-key set to undefined which would overwrite the base cache.
I added two tests (sync/async) to demonstrate and test this problem

test/integration/liquid/cache.ts
mastodon0: Sorry about that, fixed

harttle:
@all-contributors please add mastodon0 for code.

PR #223, opened 2020-04-26
fix elsif spelling on tags overview in docs

aciccarello:


harttle:
Try commit with message like `docs: blabla` instead.

PR #231, opened 2020-06-06
Fix: respect parameter order when using "for ... reversed"

stedman:
With regards to the `for` tag's [reversed](https://liquidjs.com/tags/for.html#reversed) parameter, I found the current behavior of reversing a collection *after* all the other parameters execute a bit unexpected (and not very useful).

Given `{% for i in (1..8) reversed limit: 4 %}{{ i }}{% endfor %}`, I would expect the params are executed from left to right. In other words, the collection is reversed first and then limited to four items with an output of `8765`. 

Instead I get `4321`, which is what I would expect with `{% for i in (1..8) limit: 4 reversed %}{{ i }}{% endfor %}` where the collection is limited to the first four items before reversing.

My use case for this is reversing a large collection of blog posts (so they are listed as most recent first) and then sizing those results down to five posts for the home page.

This proposed change will impact existing code that uses `reversed` with additional params. If this PR seems too risky, perhaps adding a new parameter (`flipped`?) with the proposed functionality would do the trick.


harttle:
I'm afraid it's not conformant which means it'll break Jekyll sites and Shopify templates, though I absolutely like this point. 

Try file an issue on https://github.com/Shopify/liquid instead.

stedman:
@harttle thanks for the quick feedback! It looks like [this was proposed 5+ years ago](https://github.com/Shopify/liquid/pull/456) but fizzled out.

[I added my voice to the issue](https://github.com/Shopify/liquid/pull/456#issuecomment-640286644) with yet another suggestion to extend `reversed` with an optional value to achieve the same goal: `{% for i in (1..8) reversed:"first" limit:4 %}` outputs `8765`. We'll see what happens.

harttle:
Added `orderedFilterParameters` option to enable this feature and changed your last test case

```liquid
{% for i in (1..8) offset:2 reversed limit:3 %}{{ i }}{% endfor %}
```

expected value to from `543` to `876` because `reversed` should happen before `limit`.

And there's a strange behavior, which I consider it as a bug, in Shopify liquid that

PR #234, opened 2020-06-25
feat: setup universal browser and node builds

TrySound:
In this diff I fixed es modules distribution for bundlers.

Before we had two problems
- bundlers use "module" field to find es module build
- bundlers cannot use provided es module with node target

In this diff I addressed both of them. Now "main" and "module" fields
points to "node" specific versions. "browser" field provides aliases
with browser compatible builds.

All modern browsers support both "module" and "browser" fields.


harttle:
@all-contributors please add TrySound for code

TrySound:
It was fast, thanks!

PR #237, opened 2020-06-27
fix: enumerate Promises (e.g. in for & tablerow)

tejasmanohar:
Previously, a Promise of an array was not being enumerated in
{% for %}, for example. This is unexpected since the library
handles promises elsewhere (e.g. if you {% assign x = promiseArray %}
and then {% for v in x %}, it worked just fine.

This PR makes Promises of arrays handled by changing toEnumerable
to handle then-ables. This affects other iterators, too, e.g. tablerow,
so I put in a test for that as well.

### DIFF
--- a/src/builtin/tags/for.ts
+++ b/src/builtin/tags/for.ts
@@ -36,7 +36,7 @@ export default {
   },
   render: function * (ctx: Context, emitter: Emitter) {
     const r = this.liquid.renderer
-    let collection = toEnumerable(evalToken(this.collection, ctx))
+    let collection = yield toEnumerable(evalToken(this.collection, ctx))
 
     if (!collection.length) {
       yield r.renderTemplates(this.elseTemplates, ctx, emitter)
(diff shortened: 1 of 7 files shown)


src/builtin/tags/for.ts
```diff
@@ -36,7 +36,7 @@ export default {
   },
   render: function * (ctx: Context, emitter: Emitter) {
     const r = this.liquid.renderer
-    let collection = toEnumerable(evalToken(this.collection, ctx))
+    let collection = yield toEnumerable(evalToken(this.collection, ctx))
```
harttle, line 39: Can we yield the value instead of the `toEmumerable()`, that'll keep toEnumerable clean and not aware of promises or/and the generator trick. Like:

```javascript
let collection = toEnumerable(yield evalToken(this.collection, ctx))
```

tejasmanohar: ah sure, one sec. not that familiar with generators tbh

tejasmanohar: Updated! How's that? @harttle

tejasmanohar: View final diff, I have some dummy commits (you can squash / merge?)

harttle:
Great job! @all-contributors please add @tejasmanohar for code.

Issue #255, opened 2020-10-06
Custom isFalsy definition


amit777:
Hi! I would like isFalsy to match more closely with JavaScript's false definition. Would it be possible to add a feature to override the default isFalsy() definition as an option to the constructor?

harttle:
Not available now. I think an additional `JSTrusy` option is better cause I can't think of other cases.

amit777:
We are planning on using a Perl liquid template implementation that shares templates with nodejs. It also has a slightly different truthy implementation which is closer to JS. So I was hoping to normalize them both.

amit777:
My typescript knowledge is not so good,  I was trying to figure out a way to pass the liquid options into boolean.ts isFalsy() function.  Would you be able to provide a pointer?  I can update docs, tests and code and create pull request after.

harttle:
A simple (direct) approach would be:

- add an additional `ctx: Context` argument to `src/render/boolean.ts#isTruthy()`, and make use of `ctx.opts.isTruthy`.
- pass the context to `operatorImpls` (then passes `isTruthy()`) to  in `src/render/expression.ts#evalOperatorToken()`
- similarly, pass contexts in `src/builtin/filters/array.ts#where()` and `src/builtin/tags/if.ts#render()`

That would be enough. Let me know if you get stuck.

PR #257, opened 2020-10-07
Added jsTruthy flag to constuctor in order to enable JS style truthiness evaluation

amit777:
This is related to ticket https://github.com/harttle/liquidjs/issues/255

I was not able to make a Chinese translation of the document.  Also there are 3 unrelated tests failing that were failing before my changes. Here they are:

```
3 failing

  1) util/strftime
       Date (Year, Month, Day)
         should format %q as date suffix:

      AssertionError: expected 'th' to equal 'st'
      + expected - actual

      -th
      +st
      
      at Context.<anonymous> (test/unit/util/strftime.ts:40:30)

  2) util/strftime
       Date (Year, Month, Day)
         %j
           should take count of leap years:

      AssertionError: expected '059' to equal '060'
      + expected - actual

      -059
      +060
      
      at Context.<anonymous> (test/unit/util/strftime.ts:29:34)

  3) util/strftime
       Date (Year, Month, Day)
         %j
           should take count of leap years:

      AssertionError: expected '060' to equal '061'
      + expected - actual

      -060
      +061
      
      at Context.<anonymous> (test/unit/util/strftime.ts:33:34)

```

### DIFF
--- a/src/render/boolean.ts
+++ b/src/render/boolean.ts
@@ -1,13 +1,12 @@
 import { Context } from '../context/context'
 
 export function isTruthy (val: any, ctx: Context): boolean {
-  if ((val === null) && ctx && ctx.opts && ctx.opts.jsTruthy ) return false;
   return !isFalsy(val, ctx)
 }
 
 export function isFalsy (val: any, ctx: Context): boolean {
-  if(ctx && ctx.opts && ctx.opts.jsTruthy) {
-    return val == false
+  if(ctx.opts.jsTruthy) {
+    return !val
   }
   else {
     return val === false || undefined === val || val === null
--- a/src/render/boolean.ts
+++ b/src/render/boolean.ts
@@ -1,6 +1,15 @@
-export function isTruthy (val: any): boolean {
-  return !isFalsy(val)
+import { Context } from '../context/context'
+
+export function isTruthy (val: any, ctx: Context): boolean {
+  if ((val === null) && ctx && ctx.opts && ctx.opts.jsTruthy ) return false;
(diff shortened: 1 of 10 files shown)


src/render/boolean.ts
```diff
@@ -1,6 +1,15 @@
-export function isTruthy (val: any): boolean {
-  return !isFalsy(val)
+import { Context } from '../context/context'
+
+export function isTruthy (val: any, ctx: Context): boolean {
+  if ((val === null) && ctx && ctx.opts && ctx.opts.jsTruthy ) return false;
```
harttle, line 4: `ctx.opts.jsTruthy` is guaranteed to exist so there's no need to `ctx && ctx.opts`...

harttle, line 4: To achieve completeness and correctness, it's better to directly return `!isFalsy(val, ctx)`.

src/render/boolean.ts
```diff
@@ -1,6 +1,15 @@
...
+  return !isFalsy(val, ctx)
 }
-export function isFalsy (val: any): boolean {
-  return val === false || undefined === val || val === null
+
+export function isFalsy (val: any, ctx: Context): boolean {
+  if(ctx && ctx.opts && ctx.opts.jsTruthy) {
+    return val == false
```
harttle, line 10: I believe it should be `return !val`. An example is `[] == false` is true, which is not correct.

harttle (changes requested):

harttle:
> 3 unrelated tests failing that were failing before my changes

Fixed by b11df79

PR #260, opened 2020-10-09
Add explicit braced property access parsing

### LINKED ISSUE #259: Brace access syntax breaks in 9.x
Hey, this syntax used to work in an earlier version for picking `complex key` from the scope: `{{ ["complex key"] }}`. However, this no longer works in the latest version of liquidjs. Did this break accidentally or should it even be supported?

I checked Ruby liquid, and the syntax is supported in default configuration but not in strict mode. It seems to be the only way to access complex liquid variables with spaces and other characters so imo it would be good to support.

The exact version breaking the syntax is some version after 9.6.x. I didn't bisect the exact version because I'm more curious if this should even work or not.

wyozi:
Parses braced property accesses similar to how Ruby Liquid does it. Fixes #259

### DIFF
--- a/src/parser/tokenizer.ts
+++ b/src/parser/tokenizer.ts
@@ -207,6 +207,14 @@ export class Tokenizer {
     const value = this.readQuoted() || this.readRange()
     if (value) return value
 
+    if (this.peek() === '[') {
+      this.p++
+      const prop = this.readQuoted()
+      if (!prop) return
+      this.readTo(']')
+      return new PropertyAccessToken(prop, [], this.p)
+    }
+
     const variable = this.readWord()
     if (!variable.size()) return
 
--- a/src/parser/tokenizer.ts
+++ b/src/parser/tokenizer.ts
@@ -211,7 +211,8 @@ export class Tokenizer {
       this.p++
       const prop = this.readQuoted()
       if (!prop) return
-      this.readTo(']')
+      if (this.peek() !== ']') return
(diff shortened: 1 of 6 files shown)


src/parser/tokenizer.ts
```diff
@@ -207,6 +207,14 @@ export class Tokenizer {
...
     if (value) return value
 
+    if (this.peek() === '[') {
+      this.p++
+      const prop = this.readQuoted()
+      if (!prop) return
+      this.readTo(']')
+      return new PropertyAccessToken(prop, [], this.p)
+    }
+
```
harttle, line 217: How about `["foo"]["bar"]` or `["foo"].bar`?

We need adapt the following logic (the `PropertyAccessToken` part) to support this case (remove the `variable` constructor argument and make it the first props maybe?).

wyozi: Makes sense, although the number of potential valid syntaxes are getting a bit scary. Do you think that could be postponed into the future and this feature shipped as is (once I fix the bugs)?

harttle: > the number of potential valid syntaxes are getting a bit scary

Seems it can be done by consume the `[`, call readValue recursively(as in props brackets), and consume the `]`. This way the following cases will be supported also:

```liquid
{{["foo"].bar}}
{{[0].bar}}
{{[key].bar}}
```

PR #278, opened 2020-12-07
feat: parse and handle date timezone offsets

### LINKED ISSUE #236: Date filter converts time to local timezone
Hi there,

There is an issue with the `date` filter that uses javascript's native dates. The timezone of the javascript date is always set to the user's timezone (this is different behaviour than in ruby). For example:
```
new Date("1990-12-31T23:00:00Z")
// -> Tue Jan 01 1991 01:00:00 GMT+0200 (Eastern European Standard Time)
```
The input is parsed correctly, but it's converted to local timezone. (Note that this loses the information of the timezone of the input value). Now if I want to format the date it will show wrong year etc. depending on my timezone. For example, if your computer is set to some other timezone than utc, then the following test will fail:
```
// test/integration/builtin/filters/date.ts

it('should not convert timezones', async function () {
  return test('{{ "1990-12-31T23:00:00Z" | date: "%Y-%m-%dT%H:%M:%S"}}', '1990-12-31T23:00:00')
})
```

wyozi:
Adds timezone offset parsing (e.g. `1990-12-31T23:00:00+02:00`) and mangling the pre-format date object to the right timestamp.

I'm not sure if it even makes sense for this to go in as is, because to me the default behavior of formatting timezoned dates to the local timezone automatically makes sense most of the time in web context: e.g. blogs might usually want to show dates in user-local format. However, configuring timezone preservation in the liquid filter is also a bit weird, since for instance in node.js we will probably always want to preserve the timezone.

What about configuring this in liquid-options?

(eventually) fixes #236

### DIFF
--- a/src/util/strftime.ts
+++ b/src/util/strftime.ts
@@ -142,8 +142,8 @@ const formatCodes = {
 
 export default function (inputDate: Date, formatStr: string) {
   let d = inputDate
-  if (d instanceof TimezoneDate && d.timezoneOffset !== null) {
-    d = new Date((+d) + new Date().getTimezoneOffset() * 60 * 1000 + d.timezoneOffset * 60 * 1000)
+  if (d instanceof TimezoneDate) {
+    d = new Date((+d) + d.inputTimezoneOffset * 60 * 1000)
   }
 
   let output = ''
@@ -175,24 +175,19 @@ function format (d: Date, match: RegExpExecArray) {
 }
 
 export class TimezoneDate extends Date {
-  ISO8601_PATTERN = /^([+-]?\d{4}(?!\d{2}\b))((-?)((0[1-9]|1[0-2])(\3([12]\d|0[1-9]|3[01]))?|W([0-4]\d|5[0-2])(-?[1-7])?|(00[1-9]|0[1-9]\d|[12]\d{2}|3([0-5]\d|6[1-6])))([T\s]((([01]\d|2[0-3])((:?)[0-5]\d)?|24\:?00)([.,]\d+(?!:))?)?(\17[0-5]\d([.,]\d+)?)?([zZ]|([+-])([01]\d|2[0-3]):?([0-5]\d)?)?)?)?$/;
-  ISO8601_OFFSET_PATTERN = /^([+-])(\d{2}):(\d{2})$/;
+  ISO8601_TIMEZONE_PATTERN = /([zZ]|([+-])(\d{2}):(\d{2}))$/;
 
-  timezoneOffset: number | null = null;
+  inputTimezoneOffset: number = 0;
 
   constructor (public dateString: string) {
     super(dateString)
-    const m = dateString.match(this.ISO8601_PATTERN)
-    const tzString = m && m[21]
-    if (tzString) {
-      if (tzString === 'Z') {
-        this.timezoneOffset = 0
-      } else {
-        const m2 = tzString.match(this.ISO8601_OFFSET_PATTERN)
-        if (m2) {
-          this.timezoneOffset = (m2[1] === '+' ? 1 : -1) * (parseInt(m2[2], 10) * 60 + parseInt(m2[3], 10))
-        }
-      }
+    const m = dateString.match(this.ISO8601_TIMEZONE_PATTERN)
+    if (m && m[1] === 'Z') {
+      this.inputTimezoneOffset = this.getTimezoneOffset()
+    } else if (m && m[2] && m[3] && m[4]) {
+      const [, , sign, hours, minutes] = m;
+      const delta = (sign === '+' ? 1 : -1) * (parseInt(hours, 10) * 60 + parseInt(minutes, 10))
+      this.inputTimezoneOffset = this.getTimezoneOffset() + delta
     }
   }
 }
--- a/src/util/strftime.ts
+++ b/src/util/strftime.ts
@@ -143,7 +143,7 @@ const formatCodes = {
 export default function (inputDate: Date, formatStr: string) {
   let d = inputDate
   if (d instanceof TimezoneDate) {
-    d = new Date((+d) + d.inputTimezoneOffset * 60 * 1000)
+    d = d.getDisplayDate()
   }
 
   let output = ''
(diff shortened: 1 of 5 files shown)


wyozi:
@harttle I thought about it a bit further, and maybe it's best to go with this solution as is since it will then match what Ruby's liquid is doing. Then in future there could be either `date "<format>", "local"` or something like that to convert to the local timezone or have it be a global setting. This does result in a breaking change for people passing in dates with timezones and assuming that they automatically convert to local timezones though.

What do you think?

test/integration/builtin/filters/date.ts
```diff
@@ -12,7 +12,19 @@ describe('filters/date', function () {
...
+    return test('{{ "1991-02-22T00:00:00" | date: "%Y-%m-%dT%H:%M:%S"}}', '1991-02-22T00:00:00')
+  })
+  it('should not change the timezone between input and output', function () {
+    return test('{{ "1990-12-31T23:00:00Z" | date: "%Y-%m-%dT%H:%M:%S"}}', '1990-12-31T23:00:00')
+  })
+  it('should apply numeric offset', function () {
+    return test('{{ "1990-12-31T23:00:00+00:00" | date: "%Y-%m-%dT%H:%M:%S"}}', '1990-12-31T23:00:00')
+  })
+  it('should apply numeric offset', function () {
+    return test('{{ "1990-12-31T23:00:00-01:00" | date: "%Y-%m-%dT%H:%M:%S"}}', '1990-12-31T23:00:00')
```
wyozi, line 24: TODO mock the local timezone offset for these

wyozi: This might be quite difficult to mock actually :/ Just overriding `Date.prototype.getTimezoneOffset` doesn't affect how the constructor processes date string inputs

src/util/strftime.ts
```diff
@@ -168,3 +173,26 @@ function format (d: Date, match: RegExpExecArray) {
   if (flags['-']) padWidth = 0
   return padStart(ret, padWidth, padChar)
 }
+
+export class TimezoneDate extends Date {
+  ISO8601_PATTERN = /^([+-]?\d{4}(?!\d{2}\b))((-?)((0[1-9]|1[0-2])(\3([12]\d|0[1-9]|3[01]))?|W([0-4]\d|5[0-2])(-?[1-7])?|(00[1-9]|0[1-9]\d|[12]\d{2}|3([0-5]\d|6[1-6])))([T\s]((([01]\d|2[0-3])((:?)[0-5]\d)?|24\:?00)([.,]\d+(?!:))?)?(\17[0-5]\d([.,]\d+)?)?([zZ]|([+-])([01]\d|2[0-3]):?([0-5]\d)?)?)?)?$/;
```
harttle, line 178: Can we make it esiear since we only need the timezone part.

src/util/strftime.ts
```diff
@@ -140,7 +140,12 @@ const formatCodes = {
 };
 (formatCodes as any).h = formatCodes.b
 
-export default function (d: Date, formatStr: string) {
+export default function (inputDate: Date, formatStr: string) {
+  let d = inputDate
+  if (d instanceof TimezoneDate && d.timezoneOffset !== null) {
```
harttle, line 145: Add `new Date().getTimezoneOffset()` to `d.timezoneOffset` to eliminate the `!== null` test. It also means if d is an instance of TimezoneData, then it's always available.

src/util/strftime.ts
```diff
@@ -140,7 +140,12 @@ const formatCodes = {
...
+    d = new Date((+d) + new Date().getTimezoneOffset() * 60 * 1000 + d.timezoneOffset * 60 * 1000)
```
harttle, line 146: And it's better to be `d = d.getDate()` or ` = d.getDisplayDate()`, let TimezoneDate encapsulate the whole complexity.

src/util/strftime.ts
```diff
@@ -168,3 +173,26 @@ function format (d: Date, match: RegExpExecArray) {
   if (flags['-']) padWidth = 0
   return padStart(ret, padWidth, padChar)
 }
+
+export class TimezoneDate extends Date {
+  ISO8601_PATTERN = /^([+-]?\d{4}(?!\d{2}\b))((-?)((0[1-9]|1[0-2])(\3([12]\d|0[1-9]|3[01]))?|W([0-4]\d|5[0-2])(-?[1-7])?|(00[1-9]|0[1-9]\d|[12]\d{2}|3([0-5]\d|6[1-6])))([T\s]((([01]\d|2[0-3])((:?)[0-5]\d)?|24\:?00)([.,]\d+(?!:))?)?(\17[0-5]\d([.,]\d+)?)?([zZ]|([+-])([01]\d|2[0-3]):?([0-5]\d)?)?)?)?$/;
+  ISO8601_OFFSET_PATTERN = /^([+-])(\d{2}):(\d{2})$/;
+
+  timezoneOffset: number | null = null;
```
harttle, line 181: Then the default value can be 0 and timezoneOffset is always a number.

test/integration/builtin/filters/date.ts
harttle: If I understand correctly, the idea is: making `{{ "<datestr>" | date: "%Y-%m-%dT%H:%M:%S" }}` always equal to `<datestr>`. It's OK to not mock a different timezone in test. The travis machine is already in a different timezone, most likely.

harttle (changes requested):

harttle:
> This does result in a breaking change for people passing in dates with timezones and assuming that they automatically convert to local timezones though.

We need a liquid-option to make it not BREAKING.

test/integration/builtin/filters/date.ts
wyozi: Yeah, for most of the non-UK developers the test should be fine 😁

src/util/strftime.ts
wyozi: I eliminated the null test but made it default to zero instead of `getTimezoneOffset()` so that it still works with timezoneless inputs

Issue #280, opened 2020-12-13
Return number or boolean


Pixcell:
As far as I can tell the function parseAndRender will always returns a string, even when the output value is a number or a boolean.

I would like to use liquidjs as a variable substitution engine, in order to be able to create a template object and provide values at runtime.

I.E:

```
const template = {
 myString: "{{my-string}}",
 myNumber: "{{my-number}}",
 myBoolean: "{{my-boolean}}"
}

const rvaluesVarSubstition = async (subjects, templateContext) => {
  for (let key in subjects) {
    try {
      if (typeof subjects[key] === 'object') {
        await rvaluesVarSubstition(subjects[key], context, null)
      } else if (typeof subjects[key] === 'string') {
        subjects[key] = await liquid.parseAndRender(subjects[key], context)
      }
    } catch (e) {
      // Trace error in rendering
      console.log({ e })
      subjects[key] = e.toString()
    }
  }
  return subjects
}

const context = {
 'my-string': 'this is a test',
 'my-number': 42,
 'my-boolean': true
}

const result = await rvaluesVarSubstition(template, context)

expect(result.myNumber).to.be.a('number').and.equal(42) // Fails because it's a string equal to '42'
expect(result.myBoolean).to.be.a('boolean').and.equal(true) // Fails because it's a string equal to 'true'
```

Is there a way to prevent liquidjs from `stringifying` all the values returned ? 

Thanks!

Pixcell:
Took the liberty of creating a PR with a possible fix. Let me know if you'd be interested in such a feature, or if you need me to modify anything

Pixcell:
PR got merged. Closing the issue

PR #281, opened 2020-12-13
feat: add option for keeping variable type in output

Pixcell:
Fixes issue https://github.com/harttle/liquidjs/issues/280

### DIFF
--- a/src/builtin/tags/block.ts
+++ b/src/builtin/tags/block.ts
@@ -20,7 +20,7 @@ export default {
     const r = this.liquid.renderer
     const html = childDefined !== undefined
       ? childDefined
-      : yield r.renderTemplates(this.tpls, ctx, new Emitter(ctx.opts.keepOutputType))
+      : yield r.renderTemplates(this.tpls, ctx)
 
     if (ctx.getRegister('blockMode', BlockMode.OUTPUT) === BlockMode.STORE) {
       blocks[this.block] = html
--- a/src/builtin/tags/block.ts
+++ b/src/builtin/tags/block.ts
@@ -20,7 +20,7 @@ export default {
     const r = this.liquid.renderer
     const html = childDefined !== undefined
       ? childDefined
-      : yield r.renderTemplates(this.tpls, ctx)
+      : yield r.renderTemplates(this.tpls, ctx, new Emitter(ctx.opts.keepOutputType))
 
     if (ctx.getRegister('blockMode', BlockMode.OUTPUT) === BlockMode.STORE) {
       blocks[this.block] = html
--- a/src/render/emitter.ts
+++ b/src/render/emitter.ts
@@ -23,7 +23,7 @@ export class Emitter {
     if (this.keepOutputType === true && typeof html !== 'string' && this.html === '') {
       this.html = html
     } else {
-      this.html += html as string
+      this.html = stringify(this.html) + stringify(html)
     }
   }
 }
--- a/src/render/emitter.ts
+++ b/src/render/emitter.ts
@@ -1,3 +1,5 @@
+import { stringify, toValue } from '../util/underscore'
+
 export class Emitter {
   public html: any = '';
   public break = false;
@@ -9,6 +11,11 @@ export class Emitter {
   }
 
   public write (html: any) {
+    if (this.keepOutputType === true) {
+      html = toValue(html)
+    } else {
+      html = stringify(toValue(html))
+    }
     // This will only preserve the type if the value is isolated.
     // I.E:
     // {{ my-port }} -> 42
--- a/src/render/emitter.ts
+++ b/src/render/emitter.ts
@@ -1,9 +1,18 @@
 export class Emitter {
-  public html = '';
+  public html: any = '';
   public break = false;
   public continue = false;
+  private keepOutputType? = false;
 
-  public write (html: string) {
-    this.html += html
+  constructor (keepOutputType: boolean|undefined) {
+    this.keepOutputType = keepOutputType
+  }
+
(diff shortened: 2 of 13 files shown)


src/render/emitter.ts
```diff
@@ -9,7 +9,11 @@ export class Emitter {
   }
 
   public write (html: any) {
-    if (this.keepOutputType && typeof html !== 'string') {
+    // This will only preserve the type if the value is isolated.
+    // I.E:
+    // {{ my-port }} -> 42
+    // {{ my-host }}:{{ my-port }} -> 'host:42'
+    if (this.keepOutputType === true && typeof html !== 'string' && this.html === '') {
```
harttle, line 16: How about add a separate emitter (like SingleValueEmitter)? It can also check un-intended cases, maybe in the case `{{ my-host }}:{{ my-port }} -> 'host:42'` you need an error thrown?

src/template/output.ts
```diff
@@ -16,6 +16,10 @@ export class Output extends TemplateImpl<OutputToken> implements Template {
   }
   public * render (ctx: Context, emitter: Emitter) {
     const val = yield this.value.value(ctx)
-    emitter.write(stringify(toValue(val)))
+    if (ctx.opts.keepOutputType) {
+      emitter.write(toValue(val))
+    } else {
+      emitter.write(stringify(toValue(val)))
```
harttle, line 22: We should return the `val` as it is, and always treat it inside emitters.

> BTY, I see `stringify` already called `toValue`, see https://github.com/harttle/liquidjs/blob/236d97170519337dcb216e228d93b9208a23081e/src/util/underscore.ts#L32

src/builtin/tags/block.ts
```diff
@@ -20,7 +20,7 @@ export default {
     const r = this.liquid.renderer
     const html = childDefined !== undefined
       ? childDefined
-      : yield r.renderTemplates(this.tpls, ctx)
+      : yield r.renderTemplates(this.tpls, ctx, new Emitter(ctx.opts.keepOutputType))
```
harttle, line 23: Change the current `new Emitter()` in `renderTemplates()` (inside its parameter declaration) to creating an emitter based on `ctx.opts.keepOutputType`, so we can avoid sweeping changes on every `renderTemplates()` call (Still, I can see you're missing some calling points).

harttle (changes requested):

src/render/emitter.ts
Pixcell: I can create a new Emitter, but I still would like `{{ my-host }}:{{ my-port }} -> 'host:42'` to work without having to specify a specific filter like `| toString`.
Imho, if the value is isolated, it makes sense to keep the original type, if the value is next to others, in most scenario I would say that means you are looking at doing string concatenation.

src/template/output.ts
Pixcell: Since I am really new to liquidjs code base, I tried to make as few code change as possible in order to accomplish the output I desired.

I am however happy to refactor some of the logic and put the existing logic inside the emitter.write function

src/render/emitter.ts
Pixcell: How would you see the new Emitter working? 
If the option `keepOutputType` is `true`, it automatically creates a `SingleValueEmitter` inside `renderTemplates` ? Otherwise, it creates the default HTMLEmitter ?

src/builtin/tags/block.ts
Pixcell: Yes, I struggled to make it work with Typescript. But I jsut realised I could make the emitter param optional, and check if the value is defined at the start of the function, creating the proper emitter depending on the options if it is undefined

src/template/output.ts
Pixcell: Addressed in commit 85f30fd9dce27914357b0b547ea1cbd6bd988962

src/builtin/tags/block.ts
Pixcell: Adressed in commit 85f30fd9dce27914357b0b547ea1cbd6bd988962

src/render/emitter.ts
```diff
@@ -1,9 +1,29 @@
...
+      html = stringify(toValue(html))
+    }
+    // This will only preserve the type if the value is isolated.
+    // I.E:
+    // {{ my-port }} -> 42
+    // {{ my-host }}:{{ my-port }} -> 'host:42'
+    if (this.keepOutputType === true && typeof html !== 'string' && this.html === '') {
+      this.html = html
+    } else {
+      this.html += html as string
```
harttle, line 26: OK, it makes sense to be one Emitter. The previously set `this.html` can be a `toValue()`, which also needs `stringify`.

Pixcell:
Thanks @harttle for your feedback.
I believe I have addressed everything you wanted. Let me know if you want anything else to be modified.

Otherwise, let me know when this will be merged and released so that I can stop using my fork ;)

harttle:
> I believe I have addressed everything you wanted. Let me know if you want anything else to be modified.

There's one more https://github.com/harttle/liquidjs/pull/281/files#r542981269

> Otherwise, let me know when this will be merged and released so that I can stop using my fork ;)

It'll be released automatically when merged to master.

src/render/emitter.ts
Pixcell: No sure entirely sure I understand what you mean, but I hope this is what you were after. Updated in 68a5938

harttle:
@all-contributors please add @Pixcell  for code

Pixcell:
@harttle Thanks!

Issue #276, opened 2020-11-29
Promise support in expressions


almousa1990:
Hi,

This library support promises in context, however, this support seems to be missing when evaluating an expression

```
engine
    .parseAndRender('{%if name== blank %}it is promise, not blank!{%endif%}', {name: Promise.resolve('alice')})
    .then(console.log);
```

almousa1990:
Was able to resolve the issue by modifying Expression class, i am not sure if i am correct (yield is a concept that i am not familiar with). Anyway, this is my slightly modified version of the class:

```
export class Expression {
  private operands: any[] = []
  private postfix: IterableIterator<Token>

  public constructor (str: string) {
    const tokenizer = new Tokenizer(str)
    this.postfix = toPostfix(tokenizer.readExpression())
  }
  public * evaluate (ctx: Context): any {
    for (const token of this.postfix) {
      if (TypeGuards.isOperatorToken(token)) {
        const r = yield this.operands.pop()
        const l = yield this.operands.pop()
        const result = evalOperatorToken(token, l, r, ctx)
        this.operands.push(result)
      } else {
        this.operands.push(evalToken(token, ctx))
      }
    }
    return this.operands[0]
  }
  public * value (ctx: Context) {
    return toValue(yield this.evaluate(ctx))
  }
}
```

harttle:
This feature need substantial effort (I trie this too), essentially will change all expression evaluation to async. Also prone to performance issues.

Will leave this issue open to see how many people want this. Feel free to LIKE this.

PR #285, opened 2020-12-17
Promise support in liquidMethodMissing & expressions

### LINKED ISSUE #274: liquidMethodMissing doesn't support object promises
Hi,

In attempt to implement a drop, i am trying to return an object from liquidMethodMissing method in order to lazily load objects. 

For example:
```
class ProductsDrop extends Drop{
  constructor(api) {
      super()
  }
  async liquidMethodMissing(productHandle) {
      const product = this.api.getProductByHandle(productHandle)
      if (product) return product else return null
  }
}
```

and to call it as, for example:  {{products['product handle'].name]}}

However, it seems that liquidMethodMissing supports only string type Promises.
```
  public liquidMethodMissing (key: string): Promise<string | undefined> | string | undefined {
    return undefined
  }
```

almousa1990:
adds a promise support in liquidMethodMissing. This should extends Drop capability. Should fix #274 

Also, adds a promise support in expressions. Should address #276  

This is my first pull request, i ran the tests and all passed.

### DIFF
--- a/src/context/context.ts
+++ b/src/context/context.ts
@@ -67,9 +67,12 @@ export class Context {
   }
 }
 
-export function readProperty (obj: Scope, key: string) {
+export function readProperty (obj: Scope, key: string) : any {
   if (isNil(obj)) return obj
   obj = toLiquid(obj)
+  if (obj instanceof Promise) {
+    return obj.then(resolvedObj=>readProperty(resolvedObj, key))
+  }
   if (obj instanceof Drop) {
     if (isFunction(obj[key])) return obj[key]()
     if (obj.hasOwnProperty(key)) return obj[key]
--- a/src/context/context.ts
+++ b/src/context/context.ts
@@ -67,7 +67,7 @@ export class Context {
   }
 }
 
-export function readProperty (obj: Scope, key: string) : any {
+export function readProperty (obj: Scope, key: string): any {
   if (isNil(obj)) return obj
   obj = toLiquid(obj)
   if (obj instanceof Promise) {
--- a/src/context/context.ts
+++ b/src/context/context.ts
@@ -71,7 +71,7 @@ export function readProperty (obj: Scope, key: string) : any {
   if (isNil(obj)) return obj
   obj = toLiquid(obj)
   if (obj instanceof Promise) {
-    return obj.then(resolvedObj=>readProperty(resolvedObj, key))
+    return obj.then(resolvedObj => readProperty(resolvedObj, key))
   }
   if (obj instanceof Drop) {
     if (isFunction(obj[key])) return obj[key]()
--- a/src/drop/drop.ts
+++ b/src/drop/drop.ts
@@ -3,7 +3,7 @@ export abstract class Drop {
     return undefined
   }
 
-  public liquidMethodMissing (key: string): Promise<any | undefined> | string | undefined {
+  public liquidMethodMissing (key: string): Promise<any | undefined> | any | undefined {
     return undefined
   }
 }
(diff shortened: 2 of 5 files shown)


src/drop/drop.ts
```diff
@@ -3,7 +3,7 @@ export abstract class Drop {
     return undefined
   }
 
-  public liquidMethodMissing (key: string): Promise<string | undefined> | string | undefined {
+  public liquidMethodMissing (key: string): Promise<any | undefined> | any | undefined {
```
harttle, line 6: After changed to `any`, maybe `| undefined` is no longer nessessary?

test/integration/builtin/tags/if.ts
```diff
@@ -64,6 +64,12 @@ describe('tags/if', function () {
       const html = await liquid.parseAndRender(src, scope)
       return expect(html).to.equal('XY')
     })
+    it('should evaluate promise', async function () {
+      const src = `{%if var == 'var' %}success{%endif%}`
+      const scope = { 'var': Promise.resolve('var') }
+      const html = await liquid.parseAndRender(src, scope)
+      return expect(html).to.equal('success')
+    })
```
harttle, line 72: It would be even better if you also add a case demonstrating `{{ var }}` also works.

src/context/context.ts
```diff
@@ -67,9 +67,12 @@ export class Context {
   }
 }
 
-export function readProperty (obj: Scope, key: string) {
+export function readProperty (obj: Scope, key: string): any {
   if (isNil(obj)) return obj
   obj = toLiquid(obj)
+  if (obj instanceof Promise) {
+    return obj.then(resolvedObj => readProperty(resolvedObj, key))
```
harttle, line 74: I think we need await the `obj` when calling `readProperty` rather than inside `readProperty`, it's passed in by the caller after all.

If we also need to support `toLiquid` returning a Promise, we should make it explicit by `obj = yield toLiquid(obj)` and a corresponding test case. (I'm not sure if you want this, it's up to you).

src/context/context.ts
```diff
@@ -67,9 +67,12 @@ export class Context {
   }
```
harttle, line 67: Thess lines are possibly broken:

https://github.com/harttle/liquidjs/pull/285/files#diff-05ca53b312ab10fd3af29de11c6b119e8a3e13f8fe2ba6c9353977498da1ed41R43-R44

`.reduce` is executed syncly, and the following case will fail:

```javascript
const val = await liquid.parseAndRender('{{ foo.bar }}', { foo: Promise.resolve({bar: 'BAR'}) })
expect(var).to.equal('BAR')
```

I suggest replace the `.reduce()` with `for...of` and `yield readProperty()`.

harttle (changes requested):

src/context/context.ts
almousa1990: But don't we still need to await inside readProperty if the promise was an object with properties?

src/context/context.ts
almousa1990: Does it mean that getFromScope has to be generator function? i think this can be a breaking change.\

I tried the above test and it passed.

src/context/context.ts
harttle: I believe not, `readProperty` is not recursive and only called from `getFromScope`. We can yield the obj before pass into the `readProperty`, like:

```
readProperty(yield scope, path)
```

src/context/context.ts
harttle: It's OK to leave it alone. I'll handle this latter.

src/context/context.ts
almousa1990: Since i don't see a way to implement these changes without introducing breaking changes, i am going to close this PR.

Appreciated your time

PR #301, opened 2021-02-02
Add option for custom Operators

### LINKED ISSUE #300: Custom operators for conditionals
Hello! I have a bit of a weird feature request - we'd like to use this library for https://docs.github.com instead of [docs/liquid](https://github.com/docs/liquid), and for the most part it all works exactly the same way with a couple of minor API tweaks 🎉 However, there's one (admittedly hacky) feature that we depend on.

We have custom operators for conditionals - they aren't super complex, but we use them in thousands of markdown files. For example:

```js
Liquid.Condition.operators.ver_gt = (cond, left, right) => {
  if (!matchesVersionString(left)) return false
  if (!matchesVersionString(right)) return false

  const [leftPlan, leftRelease] = splitVersion(left)
  const [rightPlan, rightRelease] = splitVersion(right)

  if (leftPlan !== rightPlan) return false

  return semver.gt(semver.coerce(leftRelease), semver.coerce(rightRelease))
}
```

The actual behavior isn't super relevant - ultimately, it takes the same arguments as other conditionals and returns a boolean. In `docs/liquid` we're able to add properties to `Liquid.Condition.operators`; but the equivalent list isn't exposed in `liquidjs`. I believe that'd be [`/src/render/operator.ts`](https://github.com/harttle/liquidjs/blob/master/src/render/operator.ts).

What do y'all think about exposing that somewhere, potentially on the engine or as an option? Something like:

```js
new Liquid({
  operators: {
    ...liquid.operators,
    ver_gt: function () {}
  }
})
```

### Other options I've considered

Since I realize that's a bit of a weird one, I did try some other things. Filters don't work in the same way - you can't inline a conditional:

```js
const { Liquid } = require('liquidjs')
const semver = require('semver')
const engine = new Liquid()
engine.registerFilter('ver_gt', (left, right) => semver.gt(left, right))
const template = '{% if currentVersion | ver_gt: "3.0.0" %}True{% else %}False{% endif %}'

// 2.0.0 is _not_ greater than 3.0.0
engine.parseAndRender(template, { currentVersion: '2.0.0' }).then(console.log)
// -> True
```

We can do that via an `assign` on a separate line:

```js
const template = '{% assign isGt = currentVersion | ver_gt: "3.0.0" %}{% if isGt %}True{% else %}False{% endif %}'
```

But then it gets real verbose; there are about 700 cases of our custom `ver_lt` and `ver_gt` operators in the [github/docs](https://github.com/github/docs) codebase. If I'm missing something to make the first example work, please let me know!

I'd be happy to work with y'all to open a PR for exposing the operators in a way that we can tack custom ones on, or see if we can get the filter to work directly in the `if`. Let me know what you think!

JasonEtco:
This PR adds an option to the `Liquid` constructor called `operators`. It follows the `OperatorMap` type, and defaults to `Operators` (exported from `render/operator.ts`). `liquid.options.operators` is passed to `Expression` constructors, so that access to the list of operators always uses the option's value.

This doesn't fully work yet - I ran into an issue where custom operators' tokens aren't being evaluated properly. I believe this is due to the logic here:

https://github.com/harttle/liquidjs/blob/5901611af59047a1045063cb33009392c7675683/src/parser/match-operator.ts#L3-L11

This looks to be very specific to the built-in operators, and I'm not sure what the right approach is. We can't convert the custom operator keys to the above format (nested objects per character) because there could be naming collisions with the top-level keys (ex: `and`, `any` would both create top-level `a` keys). I'm not able to push a failing test, but this addition to one of the tests in the PR does fail:

```js
const engine = new Liquid({
  operators: {
    ...Operators,
    isFooBar: (l, r) => l === 'foo' && r === 'bar'
  }
})
const actual = await engine.parseAndRender('{% if "foo" isFooBar "foo" %}True{% else %}False{% endif %}')
expect(actual).to.equal('False')
```

@harttle, do you have any suggestions? Is there another way we can note that a given string is an `OperatorToken`? Also feel free to comment on any of the other changes so far in this PR.

Closes #300


harttle (commented):
As for the `trie`, it can be generated programmatically, note that this should be done only once for each Liquid instance (e.g. put it into the options parser). Like:

```javascript
function create(operators) {
    let trie = {}
    for (let [name, handler] of Object.entries(operators)) {
        let node = trie
        for (let c of name) node = node[c] || {}
        node.handler = handler
    }
    return trie
}
```

JasonEtco:
Thanks for that feedback @harttle 🙏 I've gone ahead and changed it to `defaultOperators` and `Operators`, I agree that's more clear.

I've also taken a stab at programatically generating the `trie` in 1dfcc8a3d35d3f1f04cab5d5aa239b6e650a1df7 - I ended up having to change a ton of files, largely to pass `liquid.options.operatorsTrie` to `new Tokenizer`. That is then also passed to `Expression`, and to `matchOperator`. I improved the test for custom operators and it does pass 🎉 but I'm wondering if there's a better way to pass the `trie` around, it feels pretty brittle to pass the argument to so many instances of `Tokenizer`. Would love your thoughts there!

harttle:
@all-contributors please add @JasonEtco  for code.

zeke:
Wow! Mad science. ⚡ 💧 ⚡

PR #306, opened 2021-02-05
fix: default to precedence 1 for custom operators

JasonEtco:
This fixes a small bug with custom operators introduced in #301; the `precedence` object helped to define when an operator was passed the literal value vs. a boolean. `0` is meant for `and/or`, and `1` for every other comparator - but custom operators would receive a value of `undefined`, which evaluated to falsy. This meant that custom operators wouldn't get the expected arguments:

```
{% if "foo" isFooBar "bar" or "foo" == "bar" %}True{% else %}False{% endif %}
```

`isFooBar` would receive:

```js
function isFooBar (l, r) {
  console.log(l, r)
  // => "foo", false
}
```

This PR falls back to a precedence of `1` for any token that isn't explicitly hard-coded (the default operators). Technically, this limits the functionality of custom operators but I think that's a reasonable tradeoff; I can't think of a kind of custom operator that would _need_ a precedence of `0`.

I've added a test for this behavior - it fails without the change to `src/tokens/operator-token.ts`, and passes with it 🎉

### DIFF
--- a/src/tokens/operator-token.ts
+++ b/src/tokens/operator-token.ts
@@ -25,6 +25,7 @@ export class OperatorToken extends Token {
     this.operator = this.getText()
   }
   getPrecedence () {
-    return precedence[this.getText()]
+    const key = this.getText()
+    return key in precedence ? precedence[key] : 1
   }
 }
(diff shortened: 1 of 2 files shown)


harttle:
Great work!

JasonEtco:
Thanks for the quick merge!

PR #333, opened 2021-04-13
Demo updated

kayuapi:


demo/typescript/todolist.liquid
```diff
@@ -1,4 +1,5 @@
-{%header content: "welcome to liquid" | capitalize%}
+{%assign title = title%}
```
harttle, line 1: what is this line for?

harttle (changes requested):

demo/typescript/todolist.liquid
kayuapi: oops, my bad! that line is unnecessary

harttle:
@all-contributors please add @kayuapi  for doc

PR #343, opened 2021-05-06
feat: when tag with multiple values

mems:
Add support for when tag with multiple values, like described in the [Shopify's liquid documentation](https://shopify.github.io/liquid/tags/control-flow/#casewhen):

Input:

```
{% assign handle = "cake" %}
{% case handle %}
  {% when "cake" %}
     This is a cake
  {% when "cookie", "biscuit" %}
     This is a cookie
  {% else %}
     This is not a cake nor a cookie
{% endcase %}
```

Output:

```
This is a cake
```

### DIFF
--- a/docs/source/tags/case.md
+++ b/docs/source/tags/case.md
@@ -12,7 +12,7 @@ Input
 {% case handle %}
   {% when "cake" %}
      This is a cake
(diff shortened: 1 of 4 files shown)


harttle:
@all-contributors please add @mems for code.

PR #366, opened 2021-07-05
Update layout.md

ilhamdev0:
Fix typo in docs

### DIFF
--- a/docs/source/tags/layout.md
+++ b/docs/source/tags/layout.md
@@ -12,7 +12,7 @@ Renders current template inside a layout template from the template [roots][root
 {% layout 'footer.liquid' %}
 ```
 
-Whe the [extname][extname] option is set, the above `.liquid` extension can be omitted and writes:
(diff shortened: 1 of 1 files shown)


docs/source/tags/layout.md
```diff
@@ -31,7 +31,7 @@ Variables defined in current template can be passed to a the layout template by
 
 ## Blocks
 
-The layout file can contain multiple `block`s which will be populated by the child template (the caller). For example we have a `default-layout.liquid` file with the following contents:
```
harttle, line 34: This is intended. `block` is a tag name, it means multiple `block` tags.

harttle (changes requested):
If `block`s  seems confusing, maybe "`block` tags" is better?

docs/source/tags/layout.md
ilhamdev0: ok that's make sense, do I need to undo those changes?

PR #372, opened 2021-07-26
V10

harttle:
Main thread for milestone [10.0.0 developer friendly, shopify friendly](https://github.com/harttle/liquidjs/milestone/3)


harttle:
Decided use v9.26 to make these improvements and try my best to keep backward-compatible.

PR #437, opened 2022-01-12
fix: remove limit on operator char length

ameyaapte1:
The matchOperator function has a default value of str.length for the end parameter


harttle (approved):

harttle:
@all-contributors please add @ameyaapte1 for code

NFTVCF (changes requested):
Why

PR #456, opened 2022-01-25
fix: where-filter null handling improved

mastodon0:
The where-filter didn't handle nulls/undefineds properly. The comparison of NullDrop and null would fail and the list would be empty.


mastodon0:
Sorry, i fumbled my commit and will recreate the pull request

harttle:
Just reset and commit again, force push to the same branch will update this pr.

PR #457, opened 2022-01-25
fix: where-filter null handling improved

mastodon0:
The where-filter didn't handle nulls/undefineds properly. The comparison of NullDrop and null would fail and the list would be empty.

### DIFF
--- a/src/builtin/filters/array.ts
+++ b/src/builtin/filters/array.ts
@@ -3,6 +3,7 @@ import { toArray } from '../../util/collection'
 import { isTruthy } from '../../render/boolean'
 import { FilterImpl } from '../../template/filter/filter-impl'
 import { Scope } from '../../context/scope'
+import { isComparable } from '../../drop/comparable'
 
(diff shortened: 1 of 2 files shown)


src/builtin/filters/array.ts
```diff
@@ -40,7 +41,9 @@ export function slice<T> (v: T[], begin: number, length = 1): T[] {
 export function where<T extends object> (this: FilterImpl, arr: T[], property: string, expected?: any): T[] {
   return toArray(arr).filter(obj => {
     const value = this.context.getFromScope(obj, String(property).split('.'))
-    return expected === undefined ? isTruthy(value, this.context) : value === expected
+    if (expected === undefined) return isTruthy(value, this.context)
+    if (isComparable(expected)) return expected.equals(value)
+    return value === expected
```
harttle, line 46: Maybe we'll need a util `equals()` to do this. I'll check whether there's another case for this.

harttle (approved):

PR #514, opened 2022-07-02
feat: inline comment tag

jg-rp:
This pull request implements the inline comment tag, as found in [Ruby Liquid version 5.4.0](https://github.com/Shopify/liquid/pull/1498). At the time of writing, Ruby Liquid version 5.4.0 is "unreleased", but inline comments [are already available in Shopify store templates](https://github.com/Shopify/liquid/pull/1498#issuecomment-1143594442).

Inline comments look like this:
```
{% # This is a comment %}
{%-
  # Comments can span multiple lines,
  # but every line must start with a hash.
-%}
```

And inside `liquid` tags:
```
{% liquid
  # This is a comment
  echo "Hello"
%}
```
You might also be interested in [this modified Prism language definition for Liquid](https://github.com/jg-rp/liquid/blob/docs/src/theme/prism-liquid-custom.js).

### DIFF
--- a/src/parser/tokenizer.ts
+++ b/src/parser/tokenizer.ts
@@ -235,13 +235,9 @@ export class Tokenizer {
 
   readTagName (): string {
     this.skipBlank()
-    const begin = this.p
     // Handle inline comment tags
-    if (this.input[this.p] === '#') {
-      return this.input.slice(begin, ++this.p)
-    }
-    while (this.peekType() & IDENTIFIER) ++this.p
-    return this.input.slice(begin, this.p)
+    if (this.input[this.p] === '#') return this.input.slice(this.p, ++this.p)
+    return this.readIdentifier().getText()
   }
 
   readHashes (jekyllStyle?: boolean) {
--- a/src/parser/tokenizer.ts
+++ b/src/parser/tokenizer.ts
@@ -233,6 +233,17 @@ export class Tokenizer {
     return new IdentifierToken(this.input, begin, this.p, this.file)
   }
 
+  readTagName (): string {
+    this.skipBlank()
+    const begin = this.p
+    // Handle inline comment tags
+    if (this.input[this.p] === '#') {
+      return this.input.slice(begin, ++this.p)
+    }
+    while (this.peekType() & IDENTIFIER) ++this.p
+    return this.input.slice(begin, this.p)
+  }
+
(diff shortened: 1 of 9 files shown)


src/parser/tokenizer.ts
```diff
@@ -234,6 +233,17 @@ export class Tokenizer {
     return new IdentifierToken(this.input, begin, this.p, this.file)
   }
 
+  readTagName (): string {
+    this.skipBlank()
+    const begin = this.p
+    // Handle inline comment tags
+    if (this.input[this.p] === '#') {
+      return this.input.slice(begin, ++this.p)
+    }
```
harttle, line 242: can we call `readIdentifier` here?

harttle (approved):

harttle:
We'll need to add [docs for this tag](https://github.com/harttle/liquidjs/tree/master/docs/source/tags) and [website sidebar](https://github.com/harttle/liquidjs/blob/4a0186d62a0c9d16afe75d26861e575cc5f51228/docs/source/_data/sidebar.yml#L79). I can help update the zh-cn translation.

src/parser/tokenizer.ts
jg-rp: Can do, after the check for `#`. '#' being an invalid character for identifiers.

I was expecting a minor performance improvement by avoiding the creation of an `IdentifierToken` for every tag name, as `readIdentifier` does. Both `TagToken` and `LiquidTagToken` were immediately calling `getText()` on the `IdentifierToken` to get the string. I've not benchmarked it though.

jg-rp:
> We'll need to add [docs for this tag](https://github.com/harttle/liquidjs/tree/master/docs/source/tags) and [website sidebar](https://github.com/harttle/liquidjs/blob/4a0186d62a0c9d16afe75d26861e575cc5f51228/docs/source/_data/sidebar.yml#L79). I can help update the zh-cn translation.

I'll have a go at docs for this tag within the next day or two.

src/parser/tokenizer.ts
jg-rp: Benchmarking showed no discernible performance improvement. Refactored accordingly in 200fb36.

harttle: Makes sense, I think it’s OK either way.

Issue #515, opened 2022-07-05
for tag does not support valueOf recognition in Drop


maiyuanrdt:
Shopify Code example:
```html
<ul> 
  {%- for field in form.errors -%}
    <li>
      {%- if field == 'form' -%}
        {{ form.errors.messages[field] }}
      {%- else -%}
        <a href="#RegisterForm-{{ field }}">
          {{ form.errors.translated_fields[field] | capitalize }}
          {{ form.errors.messages[field] }}
        </a>
      {%- endif -%}
    </li>
  {%- endfor -%}
</ul>
```
The above is the liquid code structure that outputs the error information in the form form in Shopify, in which for is used to traverse form.errors and use it as an index key.
The result of execution in Shopify is:
![origin_img_v2_ac6c3940-e11b-42e2-b06f-03de03104deg](https://user-images.githubusercontent.com/88075366/177267162-c5bc9bcc-1f49-4117-8895-b1e6f588d717.jpg)
and
![origin_img_v2_a4e5db1d-0d8a-43cb-b204-320769d92d8g](https://user-images.githubusercontent.com/88075366/177267198-0c8599b5-229d-46dc-a658-9240ebc01a67.jpg)

-------------------------------------------------------------problem lies in------------------------------------------------------------
Using liquidjs for simulation adaptation, it was found that the valueOf in Drop was not responded to.
```javascript
class ErrorsDrop extends Drop {
    constructor() {
        super()
        this.password = 'pppp'
        this.email = 'eeee'
    }

    valueOf() {
        return ['xxx', 'ccc']
    }
}
const errors = new ErrorsDrop()
liquid.parseAndRender('-- This {{ name }}: {{errors}} -- {% for field in errors %} ={{ field }}= {% endfor %}', {
    name: 'valueOf',
    errors
})
```
I defined an object named ErrorsDrop, assigned it to the errors variable, and traversed the errors through the for tag, and got the following results:
```
-- This valueOf: xxxccc --  =passwordpppp=  =emaileeee= 
```
Personally think that when for in is followed by a Drop object, it should respond to the result after its valueOf(). Otherwise, writing similar to Shopify will not take effect

harttle:
Try adding json filter as you’re in Shopify template 

```liquid
This {{ name }}: {{errors|json}} -- {% for field in errors %} ={{ field|json }}= {% endfor %}', {
    name: 'valueOf',
```

The way LiquidJS iterate objects can be different. I can try fix it as a bug.

[...]

PR #517, opened 2022-07-09
feat: support iteration protocols

jg-rp:
This pull request adds support for looping over objects implementing JavaScript's [iteration protocols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols) in Liquid `for`, `render` and `tablerow` tags. It is also a potential solution to issue #515, without changing the behaviour of `Drop.valueOf()`.

Note that this means we can loop over [Maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) and [Sets](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), as well as user-defined iterables.

**Example**
```js
const { Liquid } = require("liquidjs");
const engine = new Liquid();

class SomeIterable {
  * [Symbol.iterator]() {
    yield 'a'
    yield 'b'
    yield 'c'
  }
}

engine
  .parseAndRender(
    "{% for i in someIterable %}{{ i }}{{ endfor }}",
    {someIterable: new SomeIterable()})
  .then(console.log);
```

**Output**
```text
abc
```

Or an iterable `Drop` that can display a user friendly string if output directly.

**Example**
```js
const { Liquid, Drop } = require("liquidjs");
const engine = new Liquid();

class SomeIterableDrop extends Drop {
  constructor() {
    this.someArray = [1,2,3];
  }

  [Symbol.iterator]() {
    return this.someArray[Symbol.iterator]();
  }

  valueOf() {
    return "SomeIterableDrop";
  }
}

engine
  .parseAndRender(
    "{{ someDrop }}: {% for i in someDrop %}{{ i }} {{ endfor }}",
    {someDrop: new SomeIterableDrop()})
  .then(console.log);
```

**Output**
```text
SomeIterableDrop: 1 2 3
```

### DIFF
--- a/src/util/collection.ts
+++ b/src/util/collection.ts
@@ -1,8 +1,9 @@
-import { isNil, isString, isObject, isArray } from './underscore'
+import { isNil, isString, isObject, isArray, isIterable } from './underscore'
 
 export function toEnumerable (val: any) {
   if (isArray(val)) return val
   if (isString(val) && val.length > 0) return [val]
+  if (isIterable(val)) return Array.from(val)
   if (isObject(val)) return Object.keys(val).map((key) => [key, val[key]])
   return []
 }
--- a/src/util/underscore.ts
+++ b/src/util/underscore.ts
@@ -60,6 +60,10 @@ export function isArray (value: any): value is any[] {
   return toString.call(value) === '[object Array]'
 }
 
+export function isIterable (value: any): value is Iterable<any> {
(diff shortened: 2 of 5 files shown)


jg-rp:
A similar approach with `toArray()` might also be appropriate, adding support for iterables with filters that operate on arrays. If indeed this is a direction you would like to go.

PR #572, opened 2022-12-18
feat: forloop's parentloop

jg-rp:
This pull request implements the `parentloop` property of a `for` tag's `forloop` drop. `forloop.parentloop` was added to [Shopify/liquid version 4.0](https://github.com/Shopify/liquid/blob/master/History.md#400--2016-12-14--branch-4-0-stable).

Quoting from the Shopify/liquid [docs](https://shopify.github.io/liquid/tags/iteration/), `forloop.parentloop` is:

> The parent forloop object. If the current for loop isn’t nested inside another for loop, then nil is returned.

### DIFF
--- a/src/context/context.ts
+++ b/src/context/context.ts
@@ -39,8 +39,8 @@ export class Context {
     this.environments = env
     this.strictVariables = renderOptions.strictVariables ?? this.opts.strictVariables
   }
-  public getRegister (key: string, defaultValue?: any) {
-    return (this.registers[key] = this.registers[key] || (defaultValue ?? {}))
+  public getRegister (key: string, defaultValue = {}) {
+    return (this.registers[key] = this.registers[key] || defaultValue)
   }
   public setRegister (key: string, value: any) {
     return (this.registers[key] = value)
--- a/src/context/context.ts
+++ b/src/context/context.ts
@@ -39,8 +39,8 @@ export class Context {
     this.environments = env
     this.strictVariables = renderOptions.strictVariables ?? this.opts.strictVariables
   }
-  public getRegister (key: string) {
-    return (this.registers[key] = this.registers[key] || {})
+  public getRegister (key: string, defaultValue?: any) {
+    return (this.registers[key] = this.registers[key] || (defaultValue ?? {}))
   }
   public setRegister (key: string, value: any) {
     return (this.registers[key] = value)
--- a/src/drop/forloop-drop.ts
+++ b/src/drop/forloop-drop.ts
@@ -6,11 +6,11 @@ export class ForloopDrop extends Drop {
   public name: string
   public length: number
   public parentloop: ForloopDrop | NullDrop
-  public constructor (length: number, collection: string, variable: string, parentloop?: ForloopDrop) {
+  public constructor (length: number, collection: string, variable: string, parentloop: ForloopDrop|NullDrop = NULL) {
     super()
     this.length = length
     this.name = `${variable}-${collection}`
-    this.parentloop = parentloop ?? NULL
+    this.parentloop = parentloop
   }
   public next () {
     this.i++
--- a/src/drop/forloop-drop.ts
+++ b/src/drop/forloop-drop.ts
@@ -1,13 +1,16 @@
 import { Drop } from './drop'
+import { NULL, NullDrop } from './null-drop'
 
 export class ForloopDrop extends Drop {
   protected i = 0
   public name: string
   public length: number
-  public constructor (length: number, collection: string, variable: string) {
+  public parentloop: ForloopDrop | NullDrop
+  public constructor (length: number, collection: string, variable: string, parentloop?: ForloopDrop) {
     super()
     this.length = length
     this.name = `${variable}-${collection}`
+    this.parentloop = parentloop ?? NULL
   }
   public next () {
     this.i++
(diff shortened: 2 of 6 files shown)


test/integration/tags/for.ts
```diff
@@ -46,7 +46,7 @@ describe('tags/for', function () {
   it('should output forloop', async function () {
     const src = '{%for i in (1..1)%}{{forloop}}{%endfor%}'
     const html = await liquid.parseAndRender(src, scope)
-    return expect(html).to.equal('{"i":0,"length":1,"name":"i-(1..1)"}')
+    return expect(html).to.equal('{"i":0,"length":1,"name":"i-(1..1)","parentloop":{}}')
```
harttle, line 49: Should this contain `parentloop: {}`? It's `nil` when no parent exsit, I guess it should be `null` if it does need to appear here.

test/integration/tags/for.ts
```diff
@@ -366,4 +366,48 @@ describe('tags/for', function () {
       return expect(html).to.equal('i-someDrop i-someDrop i-someDrop ')
     })
   })
+
+  describe('parentloop', function () {
+    it('should handle no parent forloop', async function () {
+      const src = '{% for i in (1..2)%}{{ forloop.parentloop.index }}{% endfor %}'
+      const html = await liquid.parseAndRender(src, scope)
+      return expect(html).to.equal('')
```
harttle, line 374: please treat this as a question: will this throw when `strict_variables` is set in shopify/liquid?

src/context/context.ts
```diff
@@ -39,8 +39,8 @@ export class Context {
     this.environments = env
     this.strictVariables = renderOptions.strictVariables ?? this.opts.strictVariables
   }
-  public getRegister (key: string) {
-    return (this.registers[key] = this.registers[key] || {})
+  public getRegister (key: string, defaultValue?: any) {
```
harttle, line 42: I don't suppose we need change `getRegister`, could you explain?

src/context/context.ts
```diff
@@ -39,8 +39,8 @@ export class Context {
...
+    return (this.registers[key] = this.registers[key] || (defaultValue ?? {}))
```
harttle, line 43: We can eliminate the use of `defaultValue ?? {}` by setting `defaultValue = {}` in parameter line. So it's actually a *final* default value.

As mentioned before, I still don't understand why we need change this method.

src/drop/forloop-drop.ts
```diff
@@ -1,13 +1,16 @@
...
   protected i = 0
   public name: string
   public length: number
-  public constructor (length: number, collection: string, variable: string) {
+  public parentloop: ForloopDrop | NullDrop
+  public constructor (length: number, collection: string, variable: string, parentloop?: ForloopDrop) {
     super()
     this.length = length
     this.name = `${variable}-${collection}`
+    this.parentloop = parentloop ?? NULL
```
harttle, line 13: Same issue as above.

If we set `parentloop = NULL` in parameter line, we won't need `??` here.

src/drop/null-drop.ts
```diff
@@ -22,3 +22,5 @@ export class NullDrop extends Drop implements Comparable {
     return null
   }
 }
+
+export const NULL = new NullDrop()
```
harttle, line 26: nice one! could you also help change other `new NullDop` calls to `NULL`?

src/tags/for.ts
```diff
@@ -66,8 +67,14 @@ export default class extends Tag {
     }, collection)
 
     ctx.setRegister(continueKey, (hash['offset'] || 0) + collection.length)
-    const scope = { forloop: new ForloopDrop(collection.length, this.collection.getText(), this.variable) }
+
+    const forloopRegister = ctx.getRegister('forloops', [])
+    const parentloop = forloopRegister[forloopRegister.length - 1] ?? NULL
```
harttle, line 72: since `ForloopDrop` constructor already handles a `undefined` parent, here we don't need to default it to `NULL`, let it be undefined will be fine I guess.

src/context/context.ts
harttle: Sorry, now I get it. You want default to `[]` in this case. Please ignore this question.

harttle (changes requested):

harttle:
Hey @jg-rp Thank you for porting this feature. I have several questions raised in the review. I can help if you still have any douts.

test/integration/tags/for.ts
jg-rp: Of course. I've added a `toJSON()` method to `NullDrop`.

test/integration/tags/for.ts
jg-rp: Yes. I've added a test covering this case with `strictVariables` set to `true`.

src/context/context.ts
jg-rp: We could leave `getRegister` alone and store the "parentloop stack" in a property of the default object, if you prefer.

src/context/context.ts
jg-rp: Sorry. I must have been half asleep. 😞

src/drop/null-drop.ts
jg-rp: Sure. I'll do a separate PR.

harttle (approved):

PR #578, opened 2023-01-19
ownPropertyOnly not respected when passed via renderOptions

slavivanov:
When passing `ownPropertyOnly` via `renderOptions`, it is ignored, even though it's included in the RenderOptions Typescript signature.

### DIFF
--- a/src/context/context.ts
+++ b/src/context/context.ts
@@ -38,6 +38,7 @@ export class Context {
     this.globals = renderOptions.globals ?? opts.globals
     this.environments = env
     this.strictVariables = renderOptions.strictVariables ?? this.opts.strictVariables
+    this.opts.ownPropertyOnly = renderOptions.ownPropertyOnly ?? opts.ownPropertyOnly
   }
   public getRegister (key: string) {
     return (this.registers[key] = this.registers[key] || {})


src/context/context.ts
```diff
@@ -38,6 +38,7 @@ export class Context {
     this.globals = renderOptions.globals ?? opts.globals
     this.environments = env
     this.strictVariables = renderOptions.strictVariables ?? this.opts.strictVariables
+    this.opts.ownPropertyOnly = renderOptions.ownPropertyOnly ?? opts.ownPropertyOnly
```
harttle, line 41: I think we need a `this.ownPropertyOnly` instead of `this.opts`. The test case should be something like:

```typescript
const liquid = new Liquid({ ownPropertyOnly: false })

liquid.parseAndRender(tpl, {}, { ownPropertyOnly: true })
// expect using only ownproperty

liquid.parseAndRender(tpl, {})
// expect using ascendants' property
```

harttle:
@slavivanov Thank you for taking time to fix this bug. I have some comment for you.

harttle:
Merged manually, thank you @slavivanov for your work. Now it's published to NPM.

Issue #567, opened 2022-12-11
Date locale


flaviotordini:
Is it possibile to format dates with the date filter in a specific language? If not, it sounds like a good feature for a templating engine.

harttle:
We have problem in JavaScript that `Date` doesn't support format string. We'll need localizations for this purpose, I'm thinking that we can create a separate project to port a strftime lib to Liquid `date` filter. like this one: https://github.com/samsonjs/strftime#locales

pdehaan:
JavaScript does have an [`Intl.DateTimeFormat`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat), which _"enables language-sensitive date and time formatting."_

flaviotordini:
In the meantime I used this:

```js
    eleventyConfig.addFilter("localizedDate", (dateString, locale) => {
        const options = { year: 'numeric', month: 'long', day: 'numeric' };
        return new Date(dateString).toLocaleDateString(locale, options);
    });
```
The downside is that it does not support string patterns.

harttle:
Yes, you're right. We can customize locale date/time string in JavaScript, but not as flexible as in strftime.

`Intl.DateTimeFormat` and `.toLocaleDateString()` both support a format options parameter which contains something like `dateStyle`, `timeStyle`, `year`, `month`. While I can't find a way to implement strftime with it.

prassie:
@harttle would an option for global/default date format help, similar to timezoneOffset?
In current code, the default format is hard coded.
With such an option, engine users will be free to set whatever format they need, however they got it (Intl, user configured, ...).

I use the similar timezoneOffset option, to get all dates in the templates converted to the required tz.

Of course, this is for the developers who setup/init the Liquid engine, and not for the template designers.

harttle:
Makes sense. Thank you @prassie for the clarification.

harttle:
weekday and months are now locale specific, details please check: https://liquidjs.com/filters/date.html

PR #587, opened 2023-02-20
new: LiquidOptions.dateFormat to override default date format

prassie:
doc: updates related to LiquidOptions.dateFormat
doc: revamped the date filter doc for easier reference

```
const liquidjs = require("liquidjs");
console.log(new liquidjs.Liquid({ dateFormat: '%d%q of %b %Y at %I:%M %P %z'})
                        .parseAndRenderSync(`{{ '1990-12-31T23:30:28Z' | date: undefined, -330 }}`));

=> 01st of Jan 1991 at 05:00 am +0530
```

Can be used as a workaround for #567, though the date format flags need to be strftime compatible.

### DIFF
--- a/docs/source/filters/date.md
+++ b/docs/source/filters/date.md
@@ -1,77 +1,66 @@
 ---
 title: date
 ---
-
 {% since %}v1.9.1{% endsince %}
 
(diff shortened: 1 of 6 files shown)


harttle:
Thanks @prassie for updating the docs, it's surely outdated since the source code changed a lot, it indeed looks better, thank you!

PR #594, opened 2023-03-15
fix: sample FS in render-file.md

bobgubko:


harttle:
@all-contributors please add @bobgubko for code

PR #601, opened 2023-03-21
docs: Update remove.md

mahyar-pasarzangene:
The output was wrong. Filter would replace the rain all over the text.


harttle:
@all-contributors please add @mahyar-pasarzangene for doc

PR #612, opened 2023-05-22
feat: Add support for the Jekyll sample filter

TomasHubelbauer:
See https://jekyllrb.com/docs/liquid/filters

I am sorting the array randomly and then picking the first N items or all items if there is no sample limit.

### DIFF
--- a/src/filters/array.ts
+++ b/src/filters/array.ts
@@ -88,3 +88,14 @@ export function uniq<T> (arr: T[]): T[] {
     return true
   })
 }
+
+export function sample<T> (v: T[] | string, count: number | undefined = undefined): T[] | string {
+  v = toValue(v)
+  if (isNil(v)) return []
+  if (!isArray(v)) {
+    v = stringify(v)
+    return [...v].sort(() => Math.random() ).slice(0, count).join('')
+  }
+
+  return [...v].sort(() => Math.random() ).slice(0, count)
+}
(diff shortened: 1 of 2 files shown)


src/filters/array.ts
```diff
@@ -84,3 +84,14 @@ export function uniq<T> (arr: T[]): T[] {
     return true
   })
 }
+
+export function sample<T> (v: T[] | string, count: number | undefined = undefined): T[] | string {
```
harttle, line 88: could you also add test cases for this filter?

TomasHubelbauer: Done :)

test/integration/filters/array.spec.ts
```diff
@@ -104,6 +104,20 @@ describe('filters/array', function () {
       expect(html).toBe('abc')
     })
   })
+  describe('sample', function () {
```
TomasHubelbauer, line 107: This filter relies on randomness so I didn't do anything clever, just ensured the sample size is what we expect.

harttle: makes sense.

harttle:
I find it's not compliant with Jekyll's implementation [lib/jekyll/filters.rb#L355](https://github.com/jekyll/jekyll/blob/58a1f62b2349bb477fc9999c40331cecdca577d8/lib/jekyll/filters.rb#L355). Changed a bit in https://github.com/harttle/liquidjs/commit/fcb930f0ddf2489fa74cd323f24398d7e9f7717f

- return one single element instead of an array of the element when `count=1`
- set `count=1` when not specified
- change `Math.random()` to `Math.random() - 0.5` to allow true randomness, otherwise `[1, 2]` will always be `[1, 2]`

TomasHubelbauer:
Thank you for the fixes! Good idea to go by the Jekyll source code, I will do that next time I am adding a new filter.

PR #611, opened 2023-05-22
feat: Add support for the Jekyll push filter

TomasHubelbauer:
See https://jekyllrb.com/docs/liquid/filters

PR #614, opened 2023-05-25
test: Add tests for `push`

TomasHubelbauer:
I thought `push` was broken because it doesn't work on the Playground but that's not the case. Here are the tests to prove it. Taken from the `concat` tests.

Related to #611

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,21 @@
+# Contributing
+
+1. Build everything and run tests to learn how to do both:
+  - `npm run build`
+  - `npm run test`
+
+  Tests won't at first unless you've done a build at least once.
+  Subsequent changes to tests do not need re-builds but changes to the code and
+  then tests need a build to pick up the new code by the tests.
+
+2. Make your changes and add a test for them
+  - Build after you've made your changes
+  - Run tests as per the above to validate your changes
+
+3. Create a pull request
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,21 @@
+# Contributing
+
+1. Build everything and run tests to learn how to do both:
+  - `npm run build`
```
harttle, line 4: Thank you for creating this file.

`npm run build` maybe better be replaced with `npm run build:dist` to run JS build only.

harttle (approved):

CONTRIBUTING.md
TomasHubelbauer: I do have a problem with `build:docs` not passing locally so I think it is better to recommend `build:dist`, yes. 

BTW I need to bypass commit hooks, too, because of the problem with `build:docs`. I will try to figure out what the problem with it is so that the commit hooks work locally.

TomasHubelbauer: Actually I reverted this back to the full `build` command. It runs as a part of the pre-commit hook so all contributors should get it to work. I had problems with it so I dug into what was causing them and the reason was the `sed` invocations were not written in a way that was portable to BSD `sed` used by macOS (what I am using locally). I tried to fix this in #615 but we'll see if in the process of "fixing" it I broke it for GNU sed or not.

PR #615, opened 2023-05-28
fix: sed invocations to work out of the box on macOS

TomasHubelbauer:
These fixes make the sed invocations run on macOS but I am not sure if they will break the GNU sed invocations in return… I am hoping the GitHub Actions workflow will pass and config this is now portable between macOS and Ubuntu (what GitHub Actions is using).

### DIFF
--- a/bin/build-changelog.sh
+++ b/bin/build-changelog.sh
@@ -1,10 +1,13 @@
 #!/usr/bin/env bash
 
-if [[ "$OSTYPE" == "darwin"* ]]; then
-    alias sedi="sed -i ''"
-else
-    alias sedi="sed -i"
-fi
+# Run `sed` in a way that's compatible with both macOS (BSD) and Linux (GNU)
+sedi() {
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        sed -i '' "$@"
+    else
+        sed -i "$@"
+    fi
+}
 
 cd docs
 cp ../CHANGELOG.md source/tutorials/changelog.md
--- a/bin/build-changelog.sh
+++ b/bin/build-changelog.sh
@@ -20,12 +20,7 @@ sedi \
   source/tutorials/changelog.md
 cp source/tutorials/changelog.md source/zh-cn/tutorials/changelog.md
 
-if [[ "$OSTYPE" == "darwin"* ]]; then
-  sed -i '' -e '1i\
+sedi -e '1i\
  ---\ntitle: Changelog\nauto: true\n---\n' source/tutorials/changelog.md
-  sed -i '' -e '1i\
+sedi -e '1i\
  ---\ntitle: 更新日志\nauto: true\n---\n' source/zh-cn/tutorials/changelog.md
-else
-  sed -i '1i ---\ntitle: Changelog\nauto: true\n---\n' source/tutorials/changelog.md
-  sed -i '1i ---\ntitle: 更新日志\nauto: true\n---\n' source/zh-cn/tutorials/changelog.md
-fi
--- a/bin/build-changelog.sh
+++ b/bin/build-changelog.sh
@@ -1,22 +1,26 @@
 #!/usr/bin/env bash
 
+REPLACE_START_RAW='s/{%/{% raw %}{%{% endraw %}/g'
+REPLACE_END_RAW='s/{{/{% raw %}{{{% endraw %}/g'
+ESCAPE_QUOT='1 s/"/\&quot;/g'
+ESCAPE_LT='1 s/</\&lt;/g'
+ESCAPE_GT='1 s/>/\&gt;/g'
+
 cd docs
 cp ../CHANGELOG.md source/tutorials/changelog.md
-sed -i '' \
-  -e 's/{%/{% raw %}{%{% endraw %}/g' \
-  -e 's/{{/{% raw %}{{{% endraw %}/g' \
-  -e '1 s/"/\&quot;/g' \
-  -e '1 s/</\&lt;/g' \
-  -e '1 s/>/\&gt;/g' \
-  source/tutorials/changelog.md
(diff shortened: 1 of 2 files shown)


TomasHubelbauer:
From the GitHub Actions workflow:

```
+ ./bin/build-contributors.sh
sed: can't read : No such file or directory
sed: can't read : No such file or directory
```

This is what I was worried about, my changes made this portable for BSD `sed` but in the process broke it for GNU `sed`. I will try to add OS detection and call `sed` differently for macOS.

TomasHubelbauer:
Alright I just added macOS specific branches to the shell scripts and I call `sed` differently depending on what the OS is the script is running under. Now the CI passes under Ubuntu so I don't think I broke the GNU `sed` invocations and it also works on macOS locally so BSD `sed` invocations work as well.

bin/build-changelog.sh
```diff
@@ -1,15 +1,26 @@
...
-  -e 's/{%/{% raw %}{%{% endraw %}/g' \
-  -e 's/{{/{% raw %}{{{% endraw %}/g' \
-  -e '1 s/"/\&quot;/g' \
-  -e '1 s/</\&lt;/g' \
-  -e '1 s/>/\&gt;/g' \
-  source/tutorials/changelog.md
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    sed -i '' -e "$REPLACE_START_RAW" -e "$REPLACE_END_RAW" -e "$ESCAPE_QUOT" -e "$ESCAPE_LT" -e "$ESCAPE_GT" source/tutorials/changelog.md
+else
+    sed -i -e "$REPLACE_START_RAW" -e "$REPLACE_END_RAW" -e "$ESCAPE_QUOT" -e "$ESCAPE_LT" -e "$ESCAPE_GT" source/tutorials/changelog.md
```
harttle, line 14: Could we somehow use variable or alias to make this simpler? Maybe something like

```bash
alias sedi="sed -i ''"
```

If we need to repeat all the args, I'm wondering if someone missed one of the them in future changes. Ultimately we'll need javascript build scripts but now I believe scripts are doing fine.

bin/build-contributors.sh
```diff
@@ -1,10 +1,19 @@
...
+else
+    sed -i -e "$REPLACE_README_MD" -e "$REPLACE_CONTRIBUTORS_PER_LINE" docs/.all-contributorsrc
+fi
 
 all-contributors --config docs/.all-contributorsrc generate
-sed -i 's/<br \/>.*<\/td>/<\/a><\/td>/g' docs/themes/navy/layout/partial/all-contributors.swig
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    sed -i '' -e "$REPLACE_CONTRIBUTOR_LINKS" docs/themes/navy/layout/partial/all-contributors.swig
+else
+    sed -i "$REPLACE_CONTRIBUTOR_LINKS" docs/themes/navy/layout/partial/all-contributors.swig
```
harttle, line 18: Same with above comment.

Using variables provides flexibility but I prefer less code and strateforward scripts. they're just scripts right? they're easier to understand with less lines.

harttle (approved):

bin/build-changelog.sh
TomasHubelbauer: I see try that here. Unfortunately it won't work in the case where `1i\` needs to have the newline after it to work on macOS though. But at least we can reduce the amount of duplication to just those two invocations if using an alias like you propose works.

TomasHubelbauer: @harttle is this what you meant? https://github.com/TomasHubelbauer/liquidjs/pull/1

When I try this I get this error:
> sedi: command not found

I looked around and it seems the suggested solution is to use Bash functions instead: https://linuxize.com/post/how-to-create-bash-aliases/#creating-bash-aliases-with-arguments-bash-functions

WDYT we should do here? Try using Bash functions or keep the doubled up argument lists?

harttle:
Yes, but I’m not sure why alias not working. Bash function is also good for
me.

Thank you for improving this!

bin/build-changelog.sh
TomasHubelbauer: Fixed! Using Bash functions works.

harttle (changes requested):

bin/build-changelog.sh
```diff
@@ -11,5 +20,12 @@ sed -i \
   source/tutorials/changelog.md
 cp source/tutorials/changelog.md source/zh-cn/tutorials/changelog.md
 
-sed -i '1i ---\ntitle: Changelog\nauto: true\n---\n' source/tutorials/changelog.md
-sed -i '1i ---\ntitle: 更新日志\nauto: true\n---\n' source/zh-cn/tutorials/changelog.md
+if [[ "$OSTYPE" == "darwin"* ]]; then
```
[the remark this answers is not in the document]
TomasHubelbauer: I don't think this can be updated because on macOS there needs to be a newline after `1s\` but not on Ubuntu/GNU sed. WDYT?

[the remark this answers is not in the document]
TomasHubelbauer: Looks like Ubuntu can handle the escaped newline after the `1i` command in this `sed` invocation. This is something macOS (or rather, BSD `sed`) requires and I was under the assumption that Ubuntu /GNU `sed` would have a problem with it, but the CI passed!

https://github.com/TomasHubelbauer/liquidjs/actions/runs/5158731497/jobs/9292757155

The difference is this:

Before: `-e '1i ---\ntitle: Changelog\nauto: true\n---\n' source/tutorials/changelog.md`
Now:
```
-e '1i\
 ---\ntitle: Changelog\nauto: true\n---\n' source/tutorials/changelog.md
```

@harttle Please double check this isn't a false positive and it actually still does what it is supposed to on Ubuntu. But as far as I can tell it works!

PR #617, opened 2023-05-29
chore: introduce a matrix with latest Ubuntu and macOS to test the build on macOS as well

TomasHubelbauer:
Once #615 is merged, the build script should work on macOS just as well as it does on Ubuntu and with this GitHub Actions matrix we can keep ensuring that it continues to work as it should.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -4,6 +4,10 @@ jobs:
   check:
     name: Check
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -4,6 +4,10 @@ jobs:
   check:
     name: Check
     runs-on: ubuntu-latest
```
harttle, line 6: I guess this line can be deleted

harttle (changes requested):

.github/workflows/check.yml
TomasHubelbauer: I think so, too 😌

PR #623, opened 2023-06-14
fix: strftime getSuffix works for all dates

jgarber623:
As written, the strftime module's `getSuffix` method (responsible for returning a date's ordinal string) works correctly for all dates in a month _except_ 11, 12, and 13. The existing code would return a string based on the last digit of the date. In American English, these three dates use the `th` suffix rather than `st`, `nd`, or `rd`.

This refactor/bug fix draws inspiration from [Rails' ActiveSupport ordinal inflector](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/locale/en.rb). The linked Ruby code is designed to work with any number, so the bug fix to this library is simpler.

The test suite is also updated with additional examples to verify the correctness of these changes.

### DIFF
--- a/src/util/strftime.spec.ts
+++ b/src/util/strftime.spec.ts
@@ -32,13 +32,30 @@ describe('util/strftime', function () {
       })
     })
     it('should format %q as date suffix', function () {
-      const st = new Date('2016-03-01 03:05:03')
-      const nd = new Date('2016-03-02 03:05:03')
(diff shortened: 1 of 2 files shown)


harttle:
@all-contributors please add @jgarber623 for code.

jgarber623:
@harttle Thanks for the merge! 🙌🏻

PR #637, opened 2023-08-11
fix: remove_last was eating an extra character

ebobby:
When the substring to remove was anywhere but the end of the larger string, `remove_last` would eat an extra character.


harttle:
@all-contributors please add @ebobby for code

Issue #639, opened 2023-08-15
`for` is not behaving properly (as defined by Ruby's liquid) when a value is null.


ebobby:
When rendering the following, with `strict: true`: 

```
{% assign hello = "hello,world" | split: "," | concat: null %}{% for i in hello %}{{ i }}{% endfor %}
```

It returns an error `undefined variable: i` which seems to be a bug, given that the variable _is_ defined, just happens to be null. This is a mismatch with the behavior of Shopify's liquid.

PR #640, opened 2023-08-16
fix: for throws undefined var with a null value with strictVariables.

ebobby:
Please advise if this ain't the right way to fix this. I _think_ I didn't break anything and this may catch other cases of the same problem.

Fixes: [639](https://github.com/harttle/liquidjs/issues/639)

### DIFF
--- a/src/context/context.ts
+++ b/src/context/context.ts
@@ -2,7 +2,7 @@ import { Drop } from '../drop/drop'
 import { __assign } from 'tslib'
 import { NormalizedFullOptions, defaultOptions, RenderOptions } from '../liquid-options'
 import { Scope } from './scope'
-import { isArray, isNil, isString, isFunction, toLiquid, InternalUndefinedVariableError, toValueSync } from '../util'
+import { isArray, isNil, isUndefined, isString, isFunction, toLiquid, InternalUndefinedVariableError, toValueSync } from '../util'
 
 type PropertyKey = string | number;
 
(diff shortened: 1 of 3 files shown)


src/context/context.ts
```diff
@@ -79,8 +79,9 @@ export class Context {
   public * _getFromScope (scope: unknown, paths: PropertyKey[] | string): IterableIterator<unknown> {
     if (isString(paths)) paths = paths.split('.')
     for (let i = 0; i < paths.length; i++) {
+      const previousScope = scope
       scope = yield readProperty(scope as object, paths[i], this.ownPropertyOnly)
-      if (isNil(scope) && this.strictVariables) {
+      if (this.strictVariables && isNil(scope) && !definedProperty(previousScope as object, paths[i], this.ownPropertyOnly)) {
```
harttle, line 84: Note that previousScope can be a liquid object, that means we'll need to do this in `readProperty` to make it right.

But I think implement this correctly can introduce unecessary complexity. Just check `isUndefined` instead of `isNil` serves the purpose. One downsite for this approach is if there's a defined property whose value is `undefined`, which itself doesn't make sense thus I think this cornercase can be accepted.

test/integration/tags/for.spec.ts
```diff
@@ -59,6 +59,12 @@ describe('tags/for', function () {
     const html = await liquid.parseAndRender(src, scope)
     return expect(html).toBe('str-"string"-string')
   })
+  it('should not report undefined variable on null value', async function () {
```
harttle, line 62: thank you for the test case!

harttle (changes requested):

src/context/context.ts
ebobby: Yeah, I originally did it there, but it just got more complicated. I thought about changing the check for undefined as well. It doesn't entirely scream "correct" but I think is far less common to have something be _explicitly_ undefined than being null.

ebobby: Thank you for the great work, by the way.

ebobby:
@harttle Changed as requested.

harttle (approved):

PR #644, opened 2023-08-18
fix: case should allow multiple values separated by or

ebobby:
This is supported by Shopify liquid.


harttle:
@all-contributors please add @ebobby for code.

PR #648, opened 2023-08-23
fix: sort and where bug when using `strictVariables`

ebobby:
Apply the same fix for `map` to `sort` and `where` which also rely on `_getFromScope`.


djanhjo (approved):

djanhjo:
Oops, wrong PR reviewed!

PR #664, opened 2024-01-12
chore: adjust typings dir

fupengl:
Adjust the compatible typing file path to keep it consistent with historical versions.

Now
```ts
import { Tag } from 'liquidjs/dist/src/tags/assign'
```

Old
```ts
import { Tag } from 'liquidjs/dist/tags/assign'
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "./dist/liquid.node.cjs.js": "./dist/liquid.browser.umd.js",
     "./dist/liquid.node.esm.js": "./dist/liquid.browser.esm.js"
   },
-  "types": "dist/src/index.d.ts",
+  "types": "dist/index.d.ts",
   "engines": {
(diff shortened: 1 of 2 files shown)


harttle:
Do we have real world scenarios where it's broken from some version? So we can change this patch from a chore to a fix. Otherwise we will not release the change until next fix or feature.

I think this change is good to have to reduce unnecessary hierarchy. Thanks for the change.

fupengl:
> Do we have real world scenarios where it's broken from some version? So we can change this patch from a chore to a fix. Otherwise we will not release the change until next fix or feature.
> 
> I think this change is good to have to reduce unnecessary hierarchy. Thanks for the change.

The version I used before was `10.6.1`. After upgrading to the latest version, an error occurred in the ts type.

PR #668, opened 2024-02-08
fix: Rely on equal for computing contains

santialbo:
This change allows the `contains` to be computed using the `toValue` of the items in the array.

Before this change if I had an array of "Drop" `contains` would not match against the value.

*this is not valid liquid, this is just to explain the case i'm trying to fix*
```
{{[Drop("hello"), Drop("bye")] | contains "hello"}} -> false
```
with the changes on this pr, it will return true.

### DIFF
--- a/src/render/operator.ts
+++ b/src/render/operator.ts
@@ -34,8 +34,8 @@ export const defaultOperators: Operators = {
   },
   'contains': (l: any, r: any) => {
     l = toValue(l)
-    if (l && isArray(l)) return l.some((i) => equal(i, r))
-    if (l && isString(l)) return l.indexOf(toValue(r)) > -1
+    if (isArray(l)) return l.some((i) => equal(i, r))
+    if (isString(l)) return l.indexOf(toValue(r)) > -1
     return false
   },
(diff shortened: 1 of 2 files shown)


src/render/operator.ts
```diff
@@ -34,8 +34,9 @@ export const defaultOperators: Operators = {
   },
   'contains': (l: any, r: any) => {
     l = toValue(l)
-    r = toValue(r)
-    return l && isFunction(l.indexOf) ? l.indexOf(r) > -1 : false
+    if (l && isArray(l)) return l.some((i) => equal(i, r))
```
harttle, line 37:
```suggestion
    if (isArray(l)) return l.some((i) => equal(i, r))
```

src/render/operator.ts
```diff
@@ -34,8 +34,9 @@ export const defaultOperators: Operators = {
...
+    if (l && isString(l)) return l.indexOf(toValue(r)) > -1
```
harttle, line 38:
```suggestion
    if (isString(l)) return l.indexOf(toValue(r)) > -1
```

harttle (changes requested):

harttle (approved):

Issue #667, opened 2024-02-07
is {%else if foo %} valid syntax?


amit777:
I noticed this liquid syntax is getting parsed correctly when it shouldn't (notice the space between `else` and `if` in `else if invalid`:

```
{% assign bar = "bar" -%}
{% if foo %}
FOO
{% else if invalid %}
THIS SHOULD NOT BE
{% else %}
ELSE
{% endif %}
```
It renders
```
THIS SHOULD NOT BE

ELSE
```

amit777:
I would have expected the parser to throw an error with this template about the `{% else if invalid %}`

joel-hamilton:
I drafted a PR that throws an error in the `else if` case, but I just noticed that Shopify isn't throwing an error in the same case. They just parse the 'invalid' `else if...` as an `else` and ignore following conditions:
<img width="418" alt="image" src="https://github.com/harttle/liquidjs/assets/12899024/8d2d6556-d113-413d-b369-8880c821b23a">

Should we be seeking to replicate the behaviour of the Ruby library here, or is throwing a parse error desired?

harttle:
Though we don't replicate many of error scenarios in Shopify,
I think we can leave it as it is, as it's already consistent with Shopify.

joel-hamilton:
Agreed that throwing an error isn't desired, I'll close the draft PR.

I think the real issue is that liquidjs is rendering multiple `else` blocks in the same conditional statement. I opened #670

PR #669, opened 2024-02-14
fix: throw error on 'else if' tag

joel-hamilton:
Throw an error if there is an invalid `else` tag:  #667

### DIFF
--- a/src/tokens/tag-token.ts
+++ b/src/tokens/tag-token.ts
@@ -19,6 +19,7 @@ export class TagToken extends DelimitedToken {
     this.tokenizer = new Tokenizer(input, options.operators, file, this.contentRange)
     this.name = this.tokenizer.readTagName()
     this.tokenizer.assert(this.name, `illegal tag syntax, tag name expected`)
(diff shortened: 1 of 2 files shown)


src/tokens/tag-token.ts
```diff
@@ -19,6 +19,7 @@ export class TagToken extends DelimitedToken {
     this.tokenizer = new Tokenizer(input, options.operators, file, this.contentRange)
     this.name = this.tokenizer.readTagName()
     this.tokenizer.assert(this.name, `illegal tag syntax, tag name expected`)
+    this.tokenizer.assert(this.name !== 'else' || this.contentRange[1] - this.contentRange[0] === 4, () => `unexpected token: ${this.tokenizer.snapshot()}`)
```
harttle, line 22: Let's try move this into tags/if.ts

harttle (changes requested):

src/tokens/tag-token.ts
harttle: And maybe tags/unless as well

joel-hamilton:
Closing

Issue #670, opened 2024-02-16
Sequential `else` tags all render


joel-hamilton:
When there are multiple sequential `else` tags, the content inside of each`else` branch is rendered.

[Eg](https://liquidjs.com/playground.html#PHA+SWY8L3A+CnslIGlmIGZhbHNlICV9CiAgZG9uJ3Qgc2hvdwogIHslIGVsc2UgJX0KICBzaG93CiAgeyUgZWxzZSAlfQogIGRvbid0IHNob3cKeyUgZW5kaWYgJX0KCjxwPlVubGVzczwvcD4KeyUgdW5sZXNzIHRydWUgJX0KICBkb24ndCBzaG93CiAgeyUgZWxzZSAlfQogIHNob3cKICB7JSBlbHNlICV9CiAgZG9uJ3Qgc2hvdwp7JSBlbmR1bmxlc3MgJX0KCjxwPkNhc2U8L3A+CnslIGNhc2UgIiIgJX0KICB7JSB3aGVuIGZhbHNlICV9CiAgICBkb24ndCBzaG93CiAgeyUgZWxzZSAlfQogICAgc2hvdwogIHslIGVsc2UgJX0KICAgIGRvbid0IHNob3cKeyUgZW5kY2FzZSAlfQ==,e30=):
```liquid
<p>If</p>
{% if false %}
  don't show
  {% else %}
  show
  {% else %}
  don't show
{% endif %}

<p>Unless</p>
{% unless true %}
  don't show
  {% else %}
  show
  {% else %}
  don't show
{% endunless %}

<p>Case</p>
{% case true %}
  {% when false %}
    don't show
  {% else %}
    show
  {% else %}
    don't show
{% endcase %}
```

liquidjs renders:
```
<p>If</p>

  show
  
  don't show


<p>Unless</p>

  show
  
  don't show


<p>Case</p>

    show
  
    don't show


```

Shopify has inconsistent rendering behaviour here, it renders only the first `else` branch inside `if` and `unless` conditionals, but it renders all `else` blocks inside of `case` conditionals.

<img width="922" alt="image" src="https://github.com/harttle/liquidjs/assets/12899024/4341b8da-6e4c-4ab0-a8fc-b7bf2ec8ab4d">

I believe the correct behaviour is to render *only* the first `else` block in all conditionals (`if`, `unless` and `case`). I'd be happy to put up a fix for this.

jg-rp:
[...]

PR #671, opened 2024-02-16
fix: in conditionals, don't render anything after an else branch

joel-hamilton:
This PR enforces a simple rule: don't render anything after an `else` branch.

There are several bugs currently where `if`, `unless` and `case` conditionals can have branches that occur after an `else` render, or that branches after an `else` can actually cause the `else` branch itself not to render.

Fixes: #670 and #672

## Example
```liquid
<p>If</p>
{% if false %}
  don't show
  {%- else -%}
  show
  {%- else -%}
  don't show
{% endif %}

<p>If with Elsif</p>
{% if false %}
  don't show
  {%- else -%}
  show
  {%- elsif true -%}
  don't show
{% endif %}

<p>Unless</p>
{% unless true %}
  don't show
  {%- else -%}
  show
  {%- else -%}
  don't show
{% endunless %}

<p>Case</p>
{% case true %}
  {%- when false- %}
    don't show
  {%- else -%}
    show
  {%- else -%}
    don't show
{% endcase %}

<p>Case with When</p>
{% case true -%}
  {% when false -%}
    don't show (false)
  {%- else -%}
    show
  {%- else -%}
    don't show (second else)
  {%- when true -%}
    don't show (truthy when condition after else)
  {%- when true -%}
    don't show (truthy when condition after else)
{%- endcase %}
```

**`master` Behaviour**:
```html
<p>If</p>
showdon't show

<p>If with Elsif</p>
don't show

<p>Unless</p>
showdon't show

<p>Case</p>
showdon't show

<p>Case with When</p>
don't show (truthy when condition after else)don't show (truthy when condition after else)
```

**PR Behaviour**:
```html
<p>If</p>
show

<p>If with Elsif</p>
show

<p>Unless</p>
show

<p>Case</p>
show
 
<p>Case with When</p>
show
```

### DIFF
--- a/src/tags/case.ts
+++ b/src/tags/case.ts
@@ -35,7 +35,6 @@ export default class extends Tag {
         })
       })
       .on('tag:else', () => {
-        p = []
         elseCount++
         p = this.elseTemplates
       })
--- a/src/tags/case.ts
+++ b/src/tags/case.ts
@@ -10,8 +10,13 @@ export default class extends Tag {
     this.elseTemplates = []
 
     let p: Template[] = []
+    let elseCount = 0
     const stream: ParseStream = this.liquid.parser.parseStream(remainTokens)
       .on('tag:when', (token: TagToken) => {
+        if (elseCount > 0) {
+          return
+        }
+
         p = []
 
         const values: ValueToken[] = []
@@ -29,10 +34,14 @@ export default class extends Tag {
           templates: p
         })
       })
-      .on('tag:else', () => (p = this.elseTemplates))
+      .on('tag:else', () => {
+        p = []
+        elseCount++
+        p = this.elseTemplates
+      })
       .on('tag:endcase', () => stream.stop())
       .on('template', (tpl: Template) => {
-        if (p !== this.elseTemplates || (p === this.elseTemplates && p.length === 0)) {
+        if (p !== this.elseTemplates || (p === this.elseTemplates && elseCount === 1)) {
           p.push(tpl)
         }
       })
--- a/src/tags/case.ts
+++ b/src/tags/case.ts
@@ -31,7 +31,11 @@ export default class extends Tag {
       })
       .on('tag:else', () => (p = this.elseTemplates))
       .on('tag:endcase', () => stream.stop())
-      .on('template', (tpl: Template) => p.push(tpl))
+      .on('template', (tpl: Template) => {
+        if (p !== this.elseTemplates || (p === this.elseTemplates && p.length === 0)) {
+          p.push(tpl)
+        }
+      })
       .on('end', () => {
         throw new Error(`tag ${tagToken.getText()} not closed`)
       })
--- a/src/tags/case.ts
+++ b/src/tags/case.ts
@@ -40,7 +40,7 @@ export default class extends Tag {
       })
       .on('tag:endcase', () => stream.stop())
       .on('template', (tpl: Template) => {
-        if (p !== this.elseTemplates || (p === this.elseTemplates && elseCount === 1)) {
+        if (p !== this.elseTemplates || elseCount === 1) {
           p.push(tpl)
         }
       })
--- a/src/tags/if.ts
+++ b/src/tags/if.ts
@@ -29,7 +29,7 @@ export default class extends Tag {
       })
       .on('tag:endif', function () { this.stop() })
       .on('template', (tpl: Template) => {
-        if (p !== this.elseTemplates || (p === this.elseTemplates && elseCount === 1)) {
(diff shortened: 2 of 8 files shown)


src/tags/if.ts
```diff
@@ -6,7 +6,7 @@ export default class extends Tag {
 
   constructor (tagToken: TagToken, remainTokens: TopLevelToken[], liquid: Liquid) {
     super(tagToken, remainTokens, liquid)
-    let p
+    let p: Template[] = []
```
harttle, line 9: Good job

src/tags/case.ts
```diff
@@ -31,7 +31,11 @@ export default class extends Tag {
       })
       .on('tag:else', () => (p = this.elseTemplates))
       .on('tag:endcase', () => stream.stop())
-      .on('template', (tpl: Template) => p.push(tpl))
+      .on('template', (tpl: Template) => {
+        if (p !== this.elseTemplates || (p === this.elseTemplates && p.length === 0)) {
```
harttle, line 35: This won't work when there are multiple templates:

```
{%else%}
B{{output}}A
```

src/tags/case.ts
```diff
@@ -31,7 +31,11 @@ export default class extends Tag {
       })
       .on('tag:else', () => (p = this.elseTemplates))
```
harttle, line 32: I think we can set p to [] which serves as a blackhole when `elaeTemplates` is already set. And we need initialize it as null instead of empty string. For case

```
{%else%}{%else%}we don't want this output
```

test/e2e/issues.spec.ts
```diff
@@ -469,4 +469,19 @@ describe('Issues', function () {
     const result = engine.parseAndRenderSync('{{ÜLKE}}', { ÜLKE: 'Türkiye' })
     expect(result).toEqual('Türkiye')
   })
+  it('#670 Does not render multiple `else` branches in the same if conditional', () => {
```
harttle, line 472: We can move edge cases into tags/, while preserving the case mentioned in that issue.

harttle (changes requested):

src/tags/case.ts
joel-hamilton: I used the `p = []` and an `else` counter to prevent rendering any branches that follow the `else`.

> `{%else%}{%else%}we don't want this output`

I added an empty else block to the [tests](https://github.com/joel-hamilton/liquidjs/blob/e2b22d0187916a8d31ad7772248a4668f85424d0/test/integration/tags/unless.spec.ts#L60) and it passed without changing the initialization. Let me know if you see another edge case I'm missing though.

src/tags/case.ts
```diff
@@ -29,9 +34,16 @@ export default class extends Tag {
...
       })
-      .on('tag:else', () => (p = this.elseTemplates))
+      .on('tag:else', () => {
+        elseCount++
+        p = this.elseTemplates
+      })
       .on('tag:endcase', () => stream.stop())
-      .on('template', (tpl: Template) => p.push(tpl))
+      .on('template', (tpl: Template) => {
+        if (p !== this.elseTemplates || (p === this.elseTemplates && elseCount === 1)) {
```
harttle, line 43:
```suggestion
        if (p !== this.elseTemplates || elseCount === 1) {
```

src/tags/if.ts
```diff
@@ -6,19 +6,33 @@ export default class extends Tag {
...
+        })
+      })
+      .on('tag:else', () => {
+        elseCount++
+        p = this.elseTemplates
+      })
       .on('tag:endif', function () { this.stop() })
-      .on('template', (tpl: Template) => p.push(tpl))
+      .on('template', (tpl: Template) => {
+        if (p !== this.elseTemplates || (p === this.elseTemplates && elseCount === 1)) {
```
harttle, line 32:
```suggestion
        if (p !== this.elseTemplates || elseCount === 1) {
```

src/tags/unless.ts
```diff
@@ -5,21 +5,35 @@ export default class extends Tag {
...
+        })
+      })
+      .on('tag:else', () => {
+        elseCount++
+        p = this.elseTemplates
+      })
       .on('tag:endunless', function () { this.stop() })
-      .on('template', (tpl: Template) => p.push(tpl))
+      .on('template', (tpl: Template) => {
+        if (p !== this.elseTemplates || (p === this.elseTemplates && elseCount === 1)) {
```
harttle, line 33:
```suggestion
        if (p !== this.elseTemplates || elseCount === 1) {
```

harttle:
@all-contributors please add @joel-hamilton for code

Issue #682, opened 2024-04-17
Feature request: allow `lenientIf` for multiple operands


amedve:
Currently, `lenientIf` config option only applies to statements with a single operand.
This is a different behavior from the Shopify (Ruby) liquid.

Current behavior (`strictVariables` and `lenientIf` are both set to `true`):
- `{% if nonexistent %}{% endif %}` works fine
- `{% if nonexistent == "value" %}one{% else %}two{% endif %}` returns an error

Expected:
- `{% if nonexistent == "value" %}one{% else %}two{% endif %}` to return `two`
- `{% if nonexistent != "value" %}one{% else %}two{% endif %}`  to return `one`

PR #683, opened 2024-04-17
fix: Allow `lenientIf` for multiple operands (issue #682)

amedve:
Closes https://github.com/harttle/liquidjs/issues/682


harttle:
I think it makes sense to loosen the check when lenientIf is specified, as it's still within an if tag. And I believe this change will not break things, as we're loosening it instead of tightening.

harttle:
@all-contributors please add amedve for code.

PR #698, opened 2024-05-18
fix: allow liquidMethodMissing to return any supported value type

admtnnr:
The issue is that the type of `liquidMethodMissing` is too restrictive and only allows returning `string` rather than any supported value type. I ran into this issue when trying to return a `Drop` from `liquidMethodMissing` to enable dynamic lookups along an entire accessor chain.

### DIFF
--- a/src/drop/drop.ts
+++ b/src/drop/drop.ts
@@ -1,5 +1,5 @@
 export abstract class Drop {
-  public liquidMethodMissing (key: string | number): Promise<string | undefined> | string | undefined {
+  public liquidMethodMissing (key: string | number): Promise<any | undefined> | any | undefined {
     return undefined
   }
 }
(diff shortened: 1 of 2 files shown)


src/drop/drop.ts
```diff
@@ -1,5 +1,5 @@
 export abstract class Drop {
-  public liquidMethodMissing (key: string | number): Promise<string | undefined> | string | undefined {
+  public liquidMethodMissing (key: string | number): Promise<any | undefined> | any | undefined {
```
harttle, line 2:
```suggestion
  public liquidMethodMissing (key: string | number): Promise<any> | any {
```

harttle (changes requested):

admtnnr:
@harttle I think I addressed your single piece of feedback, let me know if there's anything else you'd like changed!

harttle:
@all-contributors please add @admtnnr for code

PR #701, opened 2024-05-23
fix: isComparable full interface check

### LINKED ISSUE #700: Enhance `isComparable` function for full Comparable interface implementation
Hello, this is my first created issue, if there is something wrong just tell me! 

The current implementation of the [isComparable](https://github.com/harttle/liquidjs/blob/master/src/drop/comparable.ts#L11) function only checks if the `equals` method is present on the object. This is leading me to issues when I'm using a custom Drop class that casually also has an implementation of `equals`, but I don't want to make it an implementation of the `Comparable` interface.

When I use the operators `<`, `>`, `<=`, and `>=` the class is detected as an implementation of Comparable and tries to execute the related functions. This happens in the [operator code](https://github.com/harttle/liquidjs/blob/master/src/render/operator.ts#L16).

### Proposed Solution
Enhance the `isComparable` function to check for the complete implementation of the `Comparable` interface, ensuring that the methods `equals`, `gt`, `geq`, `lt`, and `leq` are present on the object.

I made a fork and a PR for this. I will link it right now.

### Benefits
This will ensure that objects are properly evaluated and that they will use the proper comparison functions, improving the robustness of the template evaluation for the operators `<`, `>`, `<=`, and `>=`.

Please let me know if this enhancement is acceptable or if there are any changes or additional considerations I should take into account.

GuillermoCasalCaro:
### Summary

This pull request enhances the `isComparable` function to ensure it fully implements the `Comparable` interface, which includes the methods `equals`, `gt`, `geq`, `lt`, and `leq`.

#### Enhancement of `isComparable` Function
The `isComparable` function has been improved to check for the complete implementation of the `Comparable` interface. This ensures that the function correctly identifies objects that are meant to be compared using the methods of the `Comparable` interface.

- **Modified File**: `src/drop/comparable.ts`
- **Function Modified**: `isComparable`

### Related Issues

Fixes #700

---

### DIFF
--- a/src/drop/comparable.ts
+++ b/src/drop/comparable.ts
@@ -9,5 +9,12 @@ export interface Comparable {
 }
 
 export function isComparable (arg: any): arg is Comparable {
-  return arg && isFunction(arg.equals)
+  return (
+    arg && 
+    isFunction(arg.equals) && 
+    isFunction(arg.gt) && 
+    isFunction(arg.geq) && 
+    isFunction(arg.lt) && 
+    isFunction(arg.leq)
+  );
 }
--- a/src/drop/comparable.ts
+++ b/src/drop/comparable.ts
@@ -10,11 +10,11 @@ export interface Comparable {
 
 export function isComparable (arg: any): arg is Comparable {
   return (
-    arg && 
-    isFunction(arg.equals) && 
-    isFunction(arg.gt) && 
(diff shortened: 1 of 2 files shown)


harttle:
@all-contributors please add @GuillermoCasalCaro for code

PR #705, opened 2024-06-04
fix: use drop `valueOf` when evaluated as solitary condition

admtnnr:
# Context

I noticed an inconsistency with drops where a drop's value will go through normal value resolution if used in an expression like `if drop > 10`, but it does not if the drop is the only part of the expression like `if drop` which causes it to always evaluate as truthy (because it's an object).

Shopify's Liquid implementation doesn't seem to support functionality similar to that of `valueOf` that we have in Liquid.js, so there's not prior art we can rely on there. Given if does seem to call `valueOf` when used with some type of operator I opted to make the bare drop case consistent with the existing behavior.

### DIFF
--- a/src/render/boolean.spec.ts
+++ b/src/render/boolean.spec.ts
@@ -1,5 +1,6 @@
 import { isTruthy, isFalsy } from './boolean'
 import { Context } from '../context'
+import { Drop } from '..'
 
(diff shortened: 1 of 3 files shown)


harttle:
@all-contributors please add @admtnnr for code

Issue #728, opened 2024-08-15
v10.15.0 and later do not support Node.js v14/v15 due to `performance` usage introduced in #715


uncenter:
#715 breaks support for Node.js v14 and 15 (which are [listed as supported versions](https://github.com/harttle/liquidjs/blob/ebc8daeec9b5e14e9ce8d43567cbc3bd84641924/package.json#L13-L15)) due to the introduction of code using `performance.now()` ([1](https://github.com/harttle/liquidjs/blob/ebc8daeec9b5e14e9ce8d43567cbc3bd84641924/src/context/context.ts#L46), [2](https://github.com/harttle/liquidjs/blob/ebc8daeec9b5e14e9ce8d43567cbc3bd84641924/src/render/render.ts#L19)), which is not available in the global context until Node.js v16 (previously it had to be imported manually with `const { performance } = require('perf_hooks')`.

Discovered via https://github.com/11ty/eleventy/issues/3414.

harttle:
Thanks @uncenter @vrugtehagel , the PR is merged. I'll try to create a `node-version: 14` pipeline on GithubActions so it's checked.

PR #729, opened 2024-08-15
fix: Import performance object from perf_hooks

vrugtehagel:
Resolves https://github.com/harttle/liquidjs/issues/728 by importing the `performance` object from its corresponding `perf_hooks` module.

Note that the `build-changelog.sh` script failed to run on my machine. I ended up forcing the commit with `--no-verify`. Perhaps the changelog should be added manually? Not sure if the `build-changelog.sh` script produces something that must be committed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1648,10 +1648,13 @@
       "dev": true
     },
     "@types/node": {
-      "version": "18.15.3",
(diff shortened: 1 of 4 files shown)


harttle:
@all-contributors please add @vrugtehagel for code.

PR #757, opened 2024-09-27
docs: Replace trimValue with trimOutput

Neamar:
I believe this is a typo, as the config has been called `trimOutputLeft`/`Right` for 5 years: https://github.com/harttle/liquidjs/blame/83922032b79fd678237de8786f0d75fb33e6312f/src/liquid-options.ts#L54

I'm new to this library, so I could also be missing out on something, in which case feel free to close, but the `trimValue` options do not exist on Config.


harttle:
@all-contributors please add @Neamar for doc.

PR #767, opened 2024-11-14
feat: static analysis

jg-rp:
This PR implements template static analysis and reports template variable usage.

**Example**

```js
import { Liquid, analyze } from "liquidjs";

const engine = new Liquid();
const template = engine.parse('{% if a %}{{ b }}{% endif %}');
const analysis = analyze(template);

console.log(Object.keys(analysis.variables))
// ['a', 'b']
```

For analysis to be successful, everything implementing the `Template` interface must implement the optional (optional for backwards compatibility) `node()` method, which returns a `StaticNode` object. `analyze()` then uses those nodes to traverse the template and gather information about variables.

If a custom tag is found that does not implement `node()`, a `StaticAnalysisError` is thrown.

The result of `analyze()` is an object with `variables`, `globals` and `locals` keys. All of which are objects mapping variable (or path) names to an array of `Variable` instances, one entry for each occurrence of the variable.

- `variables` is all variables, whether they are in scope or not. Including references to names such as `forloop` from the `for` tag.
- `globals` are variables that are not in scope. These could be "global" variables that are expected to be provided by the application developer, or possible mistakes from the template author.
- `locals` are template variables that are added to the template local scope using tags like `assign`, `capture` or `increment`.

Instances of `Variable` have `row`, `col` and `file` properties, plus `segments`, which is an array of strings and numbers that make up the path to the variable.

**TODO**

- [ ] more tests
- [ ] documentation


jg-rp:
Hi @harttle. Before I go any further with this, is this something you want include in LiquidJS? and do you think the approach is OK?

src/tags/for.ts
```diff
@@ -78,6 +79,30 @@ export default class extends Tag {
...
+
+    for (const v of Object.values(this.hash.hash)) {
+      if (isValueToken(v)) {
+        values.push(v)
+      }
+    }
+
+    return {
+      token: this.token,
+      values,
```
harttle, line 100: question: is it possible to get `values` by recursively iterate `.token` and `.children`? if so, can we remove exposed `values` on each tag and only retrieve it during `analyze`?

src/tags/include.ts
```diff
@@ -40,4 +42,34 @@ export default class extends Tag {
...
+
+    if (isValueToken(this.withVar)) {
+      values.push(this.withVar)
+    }
+
+    return {
+      token: this.token,
+      values, // Values from this.hash and this.file
+      children: [],
+      blockScope, // Keys from this.hash and withVar
```
harttle, line 71: same question for `blockScope` and `templateScope`, as they seem to be transient variables only needed for `analyze`, better make them computed instead of implemented in each tag.

But this issue is not big deal, I can help with it latter.

harttle:
Thank you @jg-rp for the implementation. It's a good to have feature. Only concern is hope it won't increase bundle size for browsers significantly.

src/tags/for.ts
jg-rp: We could make `analyze()` "aware" of the standard tags, and have it "know" what properties might contain values (those properties vary quite a bit). But that would not generalize to all possible custom tags (including custom versions of standard tags), so we'd still need a way for unknown tags to tell us what values they manage.

And if we took a hybrid approach, where our recursive analysis routine reads some properties directly and relies on `Node.values` for others, whenever someone wants to traverse the template for their own purposes, they too would need to include that knowledge of the built in tags.

src/tags/include.ts
jg-rp: As an example for where `blockScope` is necessary, I often use a non-standard `{% with %}` tag, inspired by [Django's template engine](https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#with). It aliases and effectively caches long variable/property paths and keeps them in scope for the duration of its block.

For static analysis to work here, the tag must populate `blockScope` with those aliases, and in a way that informs the caller that these names go out of scope after `children` have been visited.

src/tags/for.ts
harttle: Agree with hybrid approach would create more confusion. By `values they manage`, I guess you mean like arguments for a tag. It's now technically being children so we may want them be exposed by another property other than `children`. I'm not sure whether it's a good idea to expose another prop named `arguments`, is this sufficient? As in my understanding, a template consists of name, args and children.

The idea behind it is, it'll be more useful if `node` intend to expose an AST structure, which can be used more generally apart from analyzing variables.

src/tags/include.ts
harttle: Yeah, reasonable. As we'll need `blockScope` and `templateScope`, I don't think another `values` is not acceptable. Please treat my above comments as questions/ideas, I think what you're doing currently is also OK.

test/integration/static_analysis/variables.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  it('should detect local variables', () => {
+    const engine = new Liquid()
+    const template = engine.parse('{% assign a = "foo" %}{{ a }}')
+    const analysis = analyze(template)
+
+    expect(analysis).toStrictEqual({
+      variables: { a: [new Variable(['a'], { row: 1, col: 26, file: undefined })] },
+      globals: { },
+      locals: { a: [new Variable(['a'], { row: 1, col: 1, file: undefined })] }
```
harttle, line 101: I guess locals is for variables with a local scope, they can be local only (not depend on global variables), and also can be derived from global variables. In general people also want a list of access to their global variables. Like

```liquid
{% render 'product' with featured_product as product %}
// in "product.liquid"
{{product.foo}}
```

What we actually care about is `featured_product.foo`. How do you think for this scenario?

src/tags/for.ts
jg-rp: Yeah, `values` is essentially the tag's arguments. There might be one or two cases where an argument is not a value (as in an instance of the `Value` class or `ValueToken` type), but the distinction is not really important. We could rename `values` to `arguments`.

I hadn't intended nodes returned from `Template.node()` (I should rename that) to form an AST. In my head it is more of an _auxiliary tree_, specifically designed to carry template information useful for static analysis in the absence of an AST and without introducing breaking changes.

test/integration/static_analysis/variables.spec.ts
jg-rp: I'll need to think about that. A similar situation could arise from a simple `{% assign %}` tag.

```liquid
{% assign product = featured_product %}
{{ product.foo }}
```
or even

```liquid
{% assign thing = featured_product.foo %}
{{ thing.bar }}
```

harttle: There is a subtle difference: Assign can be used with filters which is considered as forming new values, while ‘as’ is just alias.
But I think it’s OK to not support such details, especially after we introduced ‘render’ to deprecate ’include’. That means template authors already have a chance to declare them explicitly.

PR #769, opened 2024-11-15
fix: allow drops in property access

santialbo:
I noticed I wasn't able to use a Drop as a property access. This PR allows using Drops as key.

### DIFF
--- a/src/context/context.ts
+++ b/src/context/context.ts
@@ -3,7 +3,7 @@ import { Drop } from '../drop/drop'
 import { __assign } from 'tslib'
 import { NormalizedFullOptions, defaultOptions, RenderOptions } from '../liquid-options'
 import { Scope } from './scope'
-import { hasOwnProperty, isArray, isNil, isUndefined, isString, isFunction, toLiquid, InternalUndefinedVariableError, toValueSync, isObject, Limiter } from '../util'
+import { hasOwnProperty, isArray, isNil, isUndefined, isString, isFunction, toLiquid, InternalUndefinedVariableError, toValueSync, isObject, Limiter, toValue } from '../util'
 
 type PropertyKey = string | number;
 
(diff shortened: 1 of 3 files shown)


src/context/context.ts
```diff
@@ -71,8 +71,8 @@ export class Context {
   public getSync (paths: PropertyKey[]): unknown {
     return toValueSync(this._get(paths))
   }
-  public * _get (paths: PropertyKey[]): IterableIterator<unknown> {
-    const scope = this.findScope(paths[0])
+  public * _get (paths: (PropertyKey | Drop)[]): IterableIterator<unknown> {
```
harttle, line 74: I think we need change the type of `PropertyKey` to include `Drop`

src/render/expression.spec.ts
```diff
@@ -50,6 +50,16 @@ describe('Expression', function () {
...
+    it('should support drops in property access', async function () {
+      class TemplateDrop extends Drop {
+        valueOf () { return 'bar' }
+      }
+      const ctx = new Context({
+        foo: { bar: 'BAR' },
+        coo: new TemplateDrop()
+      })
+      expect(await toPromise(create('foo[coo]').evaluate(ctx, false))).toBe('BAR')
+    })
```
harttle, line 62: nit: can we also add a test case under `e2e` folder or `integration` folder to demostrate the final usage in LiquidJS template?

harttle (approved):

src/context/context.ts
santialbo: I thought of that, but the issue is it used in some other places where a Drop wouldn't make sense. Also the fact that it collides with the builtin type (type PropertyKey = string | number | symbol) makes it a bit confusing sometimes.

harttle:
@all-contributors please add @santialbo as contributor for code.

PR #774, opened 2024-11-30
fix: add sideEffects=false to package.json

timvandam:
ran into some tree shaking issues


harttle:
@all-contributors please add @timvandam for code.

PR #799, opened 2025-02-11
feat: add find_index, has, and reject filters

### LINKED ISSUE #798: Add find_index, has, reject array filters
Hello hello, our friends over at shopify liquid [shipped some new collection filters recently](https://github.com/Shopify/liquid/pull/1869), available in their themes, I would assume mirroring these here would be of interest.

There are several compatibility issues whose resolution is not obvious. Most concern how `null,undefined,blank,empty` are handled when given (or not given) as the property value to these functions, will elaborate in the pull request.

brunodccarvalho:
Implement new collection filters `find_index`, `reject` and `has` to match ruby side. Fixes #798

### DIFF
--- a/docs/source/filters/where.md
+++ b/docs/source/filters/where.md
@@ -65,20 +65,6 @@ Available products:
 - Boring sneakers
 ```
 
-The `where` filter can also be used to find a single object in an array when combined with the `first` filter. For example, say you want to show off the shirt in your new fall collection.
-
-Input
-```liquid
-{% assign new_shirt = products | where: "type", "shirt" | first %}
-
-Featured product: {{ new_shirt.title }}
-```
-
-Output
-```text
-Featured product: Hawaiian print sweater vest
-```
-
 Additionally, `property` can be any valid Liquid variable expression as used in output syntax, except that the scope of this expression is within each item. For the following `products` array:
 
 ```javascript
@@ -105,9 +91,9 @@ Output
 
 ## Jekyll style
 
-{% since %}v10.19.0{% endsince %}
+{% since %}v10.21.0{% endsince %}
 
-For Liquid users migrating from Jekyll, there's a `jekyllWhere` option to mimic the behavior of Jekyll's `where` filter. This option is set to `false` by default. When enabled, if `property` is an array, the target value is matched using `Array.includes` instead of `==`, which is particularly useful for filtering tags.
+For Liquid users migrating from Jekyll, there's a `jekyllWhere` option to mimic the behavior of Jekyll's `where` filter. This option is set to `false` by default. When enabled, if `property` is an array, the target value is matched using `Array.includes` instead of `==`, which is particularly useful for excluding tags.
 
 ```javascript
 const pages = [
@@ -118,7 +104,7 @@ const pages = [
 
 Input
 ```liquid
-{% assign selected = pages | where: 'tags', "cat" %}
+{% assign selected = pages | reject: 'tags', "cat" %}
 {% for item in selected -%}
 - {{ item.title }}
 {% endfor %}
@@ -126,7 +112,7 @@ Input
 
 Output
 ```text
-Cat Food
+Dog Food
 ```
 
 [truthy]: ../tutorials/truthy-and-falsy.html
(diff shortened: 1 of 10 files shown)


brunodccarvalho:
So the first issue is an inconsistent behaviour between `where` and `find` filters when `undefined` is given as the expected value. The `where` filter will detect `undefined` and filter for truthy entries, whereas `find` simply treats `undefined` as the literal value to grep for. Of course `expected` is `undefined` when it is omitted.

```
# data = [{ name: "Adamy" }, { name: "Blake", age: 50 }], xyz is not defined
{{ data | where: "age", xyz | json }} => [Blake]
{{ data | find: "age", xyz | json }} => Adamy
```

Consider the [ruby implementation of find](https://github.com/Shopify/liquid/blob/bfe29e/lib/liquid/standardfilters.rb#L474)

From my understanding the first behaviour is roughly the (new) ruby behaviour, and the second is the jekyll behaviour. I imagine this inconsistency should be addressed, as `find` sounds like it should be functionally equivalent to `where | first`. Here `find` is changed to follow the first behaviour like the new filters. Can also consider a system option or checking `arguments.length`

brunodccarvalho:
Ruby liquid detects `null` to search for truthy values, here we detect only `undefined`. I suspect this is not a big issue.

Now `find` and `find_index` are returning `undefined` on misses while ruby wants to return `null`. This was changed on #723. It becomes functionally different after one more step, like appending `| json` or something like this which outputs `NaN` or `0` depending on what `find` returns:

```
{% assign blake = data | find: "age", 30 -%}      # wrong age
{{ blake.age | times: 2 }}
```

harttle:
Yes I agree we need find and where to be consistent, and returning null/undefined be consistent with liquid, at least in a way that it behaves same with one more step. Not sure if there’s any other consequences but we can have a try.
Checking arguments length sounds more intuitive but we need be consistent with shopify first (same template same result). We can habe a jekyllFilter toggle latter if someone has concrete cases.

brunodccarvalho:
I think we can proceed as follows for the first issue: we have all these filters implement the ruby behaviour by default, and we overload `this.context.opts.jekyllWhere` to switch them all to the jekyll behaviour. All the variable logic can be compressed into the matcher function which is shared between the filters. This is backward-compatible for `where` with `this.context.opts.jekyllWhere === false`

Don't have a strong opinion about the other two deviations

docs/source/filters/where.md
```diff
@@ -65,20 +65,6 @@ Available products:
 - Boring sneakers
 ```
 
-The `where` filter can also be used to find a single object in an array when combined with the `first` filter. For example, say you want to show off the shirt in your new fall collection.
```
harttle, line 68: I think the behavior of `| where | first` is not changed, thus this part should be remained.

docs/source/filters/where.md
```diff
@@ -105,9 +91,9 @@ Output
 
 ## Jekyll style
 
-{% since %}v10.19.0{% endsince %}
+{% since %}v10.21.0{% endsince %}
 
-For Liquid users migrating from Jekyll, there's a `jekyllWhere` option to mimic the behavior of Jekyll's `where` filter. This option is set to `false` by default. When enabled, if `property` is an array, the target value is matched using `Array.includes` instead of `==`, which is particularly useful for filtering tags.
+For Liquid users migrating from Jekyll, there's a `jekyllWhere` option to mimic the behavior of Jekyll's `where` filter. This option is set to `false` by default. When enabled, if `property` is an array, the target value is matched using `Array.includes` instead of `==`, which is particularly useful for excluding tags.
```
harttle, line 96: this file is doc for `where`, not `reject`, did you changed the wrong file?

test/integration/filters/array.spec.ts
```diff
@@ -417,6 +417,32 @@ describe('filters/array', function () {
         - Boring sneakers
```
harttle, line 417: great job! seems to be backward compatible, at least for the existing test cases.

harttle (changes requested):

docs/source/filters/where.md
brunodccarvalho: ok added it back

brunodccarvalho:
Added some minor docs for the new behaviour of the option in `where.md`

harttle:
@all-contributors please add @brunodccarvalho for code.

PR #805, opened 2025-05-10
docs: Fix formatting bug in echo.md

gurdiga:
DISCLAIMER: This may not be the proper way to approach the issue.

Although the Markdown code is proper, on the website itself it is incorrectly rendered as <code>{{\` and `}}</code> (see screenshot below).

The reason for the disclaimer above is that I’m imagining this may be an issue at the content rendering level, and my fix here is just a workaround of that issue. — Given this, I’ll not be offended if this PR of mine is rejected and closed. 🙂

<table><tr><td>
Screenshot 2025-05-10 at 08 35 51
</table>

### DIFF
--- a/docs/source/tags/echo.md
+++ b/docs/source/tags/echo.md
@@ -4,7 +4,7 @@ title: Echo
 
 {% since %}v9.31.0{% endsince %}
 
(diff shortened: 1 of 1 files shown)


harttle:
@gurdiga good catch! It’s always appreciated to get it working anyway. Does it work on your local server?

gurdiga:
I think so. 🤔 

Screenshot 2025-05-10 at 15 10 01

Issue #827, opened 2025-10-26
feat(filter): Add Shopify filters 'base64_encode' and 'base64_decode'


rosomri:
### Feature Request

This is a request to add two standard Shopify filters, `base64_encode` and `base64_decode`, to the LiquidJS core to improve Shopify compatibility.

These filters are part of the standard Shopify Liquid filter set but are not currently implemented in LiquidJS.

### Filter Specifications

**1. `base64_encode`**

  * **Description:** Encodes a string into Base64 format.
  * **Shopify Documentation:** [https://shopify.dev/docs/api/liquid/filters/base64\_encode](https://shopify.dev/docs/api/liquid/filters/base64_encode)
  * **Shopify Example:**
    ```liquid
    {{ 'one two three' | base64_encode }}
    ```
  * **Expected Output:**
    ```
    b25lIHR3byB0aHJlZQ==
    ```

**2. `base64_decode`**

  * **Description:** Decodes a Base64-formatted string.
  * **Shopify Documentation:** [https://shopify.dev/docs/api/liquid/filters/base64\_decode](https://shopify.dev/docs/api/liquid/filters/base64_decode)
  * **Shopify Example:**
    ```liquid
    {{ 'b25lIHR3byB0aHJlZQ==' | base64_decode }}
    ```
  * **Expected Output:**
    ```
    one two three
    ```

### Implementation Suggestion

These can be implemented in JavaScript using native `Buffer` methods (in Node.js) or `btoa()` / `atob()` (in browsers).

```javascript
// Node.js implementation example
const encode = (str) => Buffer.from(str, 'utf8').toString('base64');
const decode = (str) => Buffer.from(str, 'base64').toString('utf8');

// Browser implementation example
const encode = (str) => btoa(str);
const decode = (str) => atob(str);
```

A cross-compatible solution will be needed for the final implementation.

**I would like to contribute this feature.**

harttle:
Sounds good. You can use bundle options in rollup config to bundle corresponding implementation. We already have fs impl for both platforms for your reference.

rosomri:
https://github.com/harttle/liquidjs/pull/828

PR #828, opened 2025-10-26
feat(filters): Add base64_encode and base64_decode filters for Shopify compatibility

rosomri:
### Summary
This PR adds two standard Shopify filters (`base64_encode` and `base64_decode`) to LiquidJS core to improve Shopify compatibility. These filters are part of the standard Shopify Liquid filter set but were not previously implemented in LiquidJS.

https://github.com/harttle/liquidjs/issues/827

### Changes
- **New Filters**: Added `base64_encode` and `base64_decode` filters with full Shopify compatibility
- **Platform-specific Implementation**: 
  - Node.js: Uses `Buffer.from()` for optimal performance
  - Browser: Uses `btoa`/`atob` with proper Unicode handling
- **Rollup Integration**: Added bundle options to replace implementations during build
- **Comprehensive Testing**: Added tests for both Node.js and browser implementations
- **Memory Management**: Integrated with LiquidJS's memory limit system

### Implementation Details
- `src/filters/base64.ts` - Main filter implementation
- `src/filters/base64-impl.ts` - Node.js implementation using Buffer
- `src/build/base64-impl-browser.ts` - Browser implementation using btoa/atob
- `test/integration/filters/base64.spec.ts` - Integration tests
- `src/build/base64-impl-browser.spec.ts` - Browser-specific unit tests
- Updated `rollup.config.mjs` with platform-specific bundling

### Shopify Compatibility
- ✅ Matches Shopify's exact output for `{{ 'one two three' | base64_encode }}` → `b25lIHR3byB0aHJlZQ==`
- ✅ Handles Unicode characters correctly
- ✅ Supports all input types (strings, numbers, booleans)
- ✅ Proper error handling for edge cases

### Testing
- unit tests for browser implementation
- integration tests covering all scenarios
- Round-trip encoding/decoding verification
- Unicode and special character handling
- Cross-platform compatibility verified

### Breaking Changes
None

### DIFF
--- a/docs/source/filters/base64_decode.md
+++ b/docs/source/filters/base64_decode.md
@@ -2,7 +2,7 @@
 title: base64_decode
 ---
 
-{% since %}v10.23.0{% endsince %}
+{% since %}v10.24.0{% endsince %}
 
 Decodes a Base64-formatted string back to its original text.
 
--- a/docs/source/filters/base64_encode.md
+++ b/docs/source/filters/base64_encode.md
@@ -2,7 +2,7 @@
 title: base64_encode
 ---
 
-{% since %}v10.23.0{% endsince %}
+{% since %}v10.24.0{% endsince %}
 
 Encodes a string into Base64 format.
 


src/build/base64-impl-browser.ts
```diff
@@ -0,0 +1,10 @@
+
+export function base64Encode (str: string): string {
+  return btoa(String.fromCharCode(...new TextEncoder().encode(str)))
```
harttle: Try avoid TextEncoder and Uint8Array as we don’t heavily depend on them. Use escape and decodeURIComponent trick for better compatibility.

harttle (changes requested):

PR #830, opened 2025-10-30
docs: update tutorial on operators and precedence

immerrr:
This is a rather straightforward doc update:
- noticed that `not` operator was not listed on that page, added that
- fixed precedence descriptions, [looks like they got a bit out of date, too](https://github.com/harttle/liquidjs/blob/906707833e196733e592ca52fbe5eda48b71a8a6/src/tokens/operator-token.ts#L9-L20)
- updated the link to the corresponding section of shopify docs, the old one didn't work for me


harttle:
@all-contributors please add @immerrr for docs

PR #863, opened 2026-03-23
feat: Add support of inner expressions enclosed by parentheses

### LINKED ISSUE #833: Support Value Expressions as Operands in Conditional and Loop Tags
## Problem Description

LiquidJS currently does not support using value expressions (filter operations) directly as operands in conditional statements (`if`, `elsif`, `unless`) or in loop tags (`for`). This limitation requires developers to use workarounds that make templates less readable and more verbose.

## Current Limitation

The following syntax is **NOT supported**:

```liquid
{% if (strValue | size) > 10 %}
  <a href="{{person | prepend: "https://example.com/"}}">
    {{ person | capitalize }}
  </a>
{% endif %}
```

This results in a syntax error or unexpected behavior because the value expression `(strValue | size)` cannot be used directly as an operand in the conditional statement.

### Example 1: String Length Validation in Conditionals
```liquid
{% if (username | size) >= 3 %}
  Username is valid
{% elsif (username | size) == 0 %}
  Username is required
{% else %}
  Username is too short
{% endif %}
```

### Example 2: Array Size comparison
```liquid
{% if (firstArray | size) == (secondArray | size) %}
  No items found
{% else %}
  Found {{ firstArray | size }} in firstArray and {{ secondArray | size }} in secondArray
{% endif %}
```

### Example 3: String Transformation in Conditions
```liquid
{% if (email | downcase) contains "@example.com" %}
  Internal user detected
{% endif %}
```

### Example 4: Numeric Calculations
```liquid
{% if (price | times: 0.9) < 100 %}
  Discounted price is under $100
{% endif %}
```

### Example 5: Chained Filters in Conditions
```liquid
{% if (description | strip_html | size) > 50 %}
  Long description detected
{% endif %}
```

### Example 6: For Loop with Filtered Range
```liquid
{# Loop through first N items #}
{% for i in (1..items | size) %}
  Item {{ i }}
{% endfor %}
```

### Example 7: Unless with Filter Expression
```liquid
{% unless (content | strip_html | size) > 0 %}
  No content available
{% endunless %}
```

### Example 8: Case Statement with Filtered Value
```liquid
{% case (status | downcase) %}
  {% when "active" %}
    Active status
  {% when "pending" %}
    Pending status
{% endcase %}
```

### Example 9: Complex Logical Expression
```liquid
{% if (name | size) > 0 and (email | downcase) contains "@" %}
  Valid user entry
{% endif %}
```

### Example 10: Nested Filter in Loop Limit
```liquid
{% for item in items limit: (maxItems | default: 10) %}
  {{ item }}
{% endfor %}
```

## Current Workaround

To work around this limitation, you must assign the filtered value to a variable first:

```liquid
{% assign strValueSize = strValue | size %}
{% if strValueSize > 10 %}
  <a href="{{person | prepend: "https://example.com/"}}">
    {{ person | capitalize }}
  </a>
{% endif %}
```

### More Workaround Examples

```liquid
{# Example 1: String Length Validation #}
{% assign usernameLength = username | size %}
{% if usernameLength >= 3 %}
  Username is valid
{% elsif usernameLength == 0 %}
  Username is required
{% else %}
  Username is too short
{% endif %}

{# Example 2: Array Size Check #}
{% assign itemCount = items | size %}
{% if itemCount == 0 %}
  No items found
{% else %}
  Found {{ itemCount }} items
{% endif %}

{# Example 3: String Transformation #}
{% assign emailLower = email | downcase %}
{% if emailLower contains "@example.com" %}
  Internal user detected
{% endif %}

{# Example 4: Numeric Calculations #}
{% assign discountedPrice = price | times: 0.9 %}
{% if discountedPrice < 100 %}
  Discounted price is under $100
{% endif %}

{# Example 5: Chained Filters #}
{% assign cleanDescription = description | strip_html %}
{% assign descriptionLength = cleanDescription | size %}
{% if descriptionLength > 50 %}
  Long description detected
{% endif %}

{# Example 6: For Loop with Filtered Range #}
{% assign itemCount = items | size %}
{% for i in (1..itemCount) %}
  Item {{ i }}
{% endfor %}

{# Example 7: Complex Logical Expression #}
{% assign nameLength = name | size %}
{% assign emailLower = email | downcase %}
{% if nameLength > 0 and emailLower contains "@" %}
  Valid user entry
{% endif %}
```

## Why This Matters

While the workaround functions correctly, it has several significant drawbacks:

1. **Reduced Readability**: The intent is less clear when logic is split across multiple statements
2. **Increased Verbosity**: More lines of code for simple operations
3. **Variable Pollution**: Creates intermediate variables that clutter the template scope and may conflict with other variables
4. **Maintenance Burden**: More code to maintain and understand, especially for complex expressions
5. **Cognitive Load**: Developers must mentally track intermediate variables and their purpose
6. **Inconsistency**: Filters work inline in output expressions `{{ }}` but not in logic expressions `{% %}`

## Desired Behavior

Ideally, LiquidJS should support inline value expressions (filter operations) within logical operations and loop tags, similar to how they work in output statements:

```liquid
{% if (strValue | size) > 10 %}
  {# Direct filter use in condition #}
{% endif %}
```

This would make templates:
- More concise and readable
- Easier to maintain
- More intuitive for developers
- Consistent with filter usage in output contexts (`{{ }}`)
- More expressive and powerful

## Use Cases

This feature would be particularly useful in scenarios such as:

1. **Validation Logic**: Checking string lengths, array sizes, or data formats without intermediate variables
2. **Conditional Rendering**: Showing/hiding content based on processed values
3. **Access Control**: Making decisions based on transformed user data
4. **Data Formatting**: Conditional formatting based on calculated values
5. **Dynamic Loops**: Iterating with dynamically calculated limits or ranges
6. **Complex Business Logic**: Implementing multi-step conditionals with transformations
7. **Template Optimization**: Reducing template size and complexity

## Expected Behavior

Value expressions should be evaluated before being used as operands in conditionals and loops:

```liquid
{# These should all work #}
{% if (value | size) > 10 %}...{% endif %}
{% for i in (1..items | size) %}...{% endfor %}
{% unless (text | strip_html | size) == 0 %}...{% endunless %}
{% case (status | downcase) %}...{% endcase %}
```

The parentheses would indicate to the parser that the enclosed expression should be evaluated as a complete value expression before being used in the containing statement.

## Comparison with Other Templating Engines

Many other templating engines support inline expressions in conditionals:
- **Jinja2** (Python): Supports complex expressions in conditionals
- **Twig** (PHP): Allows filter operations in conditional statements
- **Nunjucks** (JavaScript): Supports inline transformations in logic blocks
- **Handlebars** (JavaScript): Allows helper functions in conditionals

Supporting this syntax would align LiquidJS with industry standards and improve developer experience.

## Proposed Solution

Allow value expressions (enclosed in parentheses or maybe not) to be evaluated and used as operands in:
- Conditional tags: `if`, `elsif`, `unless`, `case`
- Loop tags: `for` (range expressions, limit parameters)
- Any other tag that accepts value operands

This should maintain backward compatibility while adding this enhancement. The parentheses provide a clear signal to the parser that the enclosed content is a value expression that should be fully evaluated before being used.

skynetigor:
Adds support for parenthesized expressions as operands in tags, gated behind a new `groupedExpressions` option (default `false`). Closes #833.

```js
const liquid = new Liquid({ groupedExpressions: true })
```

This is a **non-standard extension** to Liquid syntax — Ruby Liquid does not support this. The feature is opt-in to preserve strict compatibility by default and can be enabled by default in a future major version.

**Backward compatibility:** Even with `groupedExpressions: true`, all existing Liquid templates continue to work identically. The change is purely additive — it only gives meaning to syntax that previously threw a parse error (`(expr | filter)` in operand position). No valid Liquid template changes behavior. Range syntax `(1..5)` is unaffected. Templates written for standard Liquid are fully compatible regardless of the flag setting.

When enabled, parentheses that don't contain range syntax (`..`) are parsed as grouped sub-expressions, allowing filter chains and logical grouping to be used directly as operands:

```liquid
{% if (username | size) >= 3 %}...{% endif %}
{% if (a | upcase) == (b | upcase) %}...{% endif %}
{% if (name | size) > 0 and (email | downcase) contains "@" %}...{% endif %}
{% unless (content | strip_html | size) == 0 %}...{% endunless %}
{% case (status | downcase) %}{% when "active" %}...{% endcase %}
{% for i in (1..(items | size)) %}...{% endfor %}
```

Existing range syntax `(1..5)` is unaffected — a lookahead scan checks for `..` at depth 1 before deciding which parser to invoke.

## Approach

- A `looksLikeRange()` lookahead in `readValue()` routes `(` to either `readRange()` or the new `readGroupedExpression()`
- `readGroupedExpression()` temporarily limits the tokenizer boundary to the matching `)` and delegates to the existing `readFilteredValue()` parser
- A new `GroupedExpressionToken` stores the parsed expression and filters; a `resolveGroupedExpressions()` utility recursively resolves them into `Value` instances when the `Liquid` instance is available
- Tags that parse values outside of `Value` (`for`, `case/when`) call the resolver in their constructors

## Test plan

- [x] Tokenizer unit tests: parsing grouped expressions, nested parens, range fallback, flag-off behavior
- [x] Expression unit tests: existing range error behavior preserved when flag is off
- [x] Integration tests for `if`, `unless`, `for`, `case/when` with `groupedExpressions: true`
- [x] `evalValueSync` API tests for direct expression evaluation


harttle (changes requested):

skynetigor:
Many thanks for the review, I'll address all the comments 👍

src/tags/for.ts
harttle:
```suggestion
  collection: ValueToken | GroupedExpressionToken
```

src/template/value.ts
```diff
@@ -1,11 +1,29 @@
...
 import { Expression } from '../render'
 import { Tokenizer } from '../parser'
-import { assert } from '../util'
-import type { FilteredValueToken } from '../tokens'
+import { assert, isGroupedExpressionToken, isRangeToken, isPropertyAccessToken } from '../util'
+import { FilteredValueToken, Token } from '../tokens'
 import type { Liquid } from '../liquid'
 import type { Context } from '../context'
 
+export function resolveGroupedExpressions (token: Token, liquid: Liquid): void {
```
harttle: I guess this should be moved to `evalGroupedExpressionToken()`, and if some structures can be created in parse time, need to be done in `GroupedExpressionToken#constructor` and consumed during `evalGroupedExpressionToken()`

harttle (changes requested):

rosomri:
Hi @harttle, Omri from @skynetigor's team - thanks for the detailed review! I'd like to confirm the approach before implementing:

### Proposed Changes

**1. Update token structure** (fixes layering violation):
```typescript
export class GroupedExpressionToken extends Token {
  public resolvedFilters?: Filter[]  // Change from resolvedValue to resolvedFilters
}
```
Tokens will store `Filter[]` (mid-level) instead of `Value` (template-layer), maintaining proper dependency direction.

**2. Evaluate at render time** (fixes eager resolution):
```typescript
function * evalGroupedExpressionToken(token, ctx, lenient) {
  assert(token.resolvedFilters, 'grouped expression filters not resolved')
  
  let val = yield token.initial.evaluate(ctx, lenient)  // Lazy evaluation
  for (const filter of token.resolvedFilters!) {
    val = yield filter.render(val, ctx)  // Apply filters lazily
  }
  return val
}
```
This follows the generator pattern used throughout liquidjs for async/sync duality.

**3. Build Filter instances during Value construction**:
```typescript
export function resolveGroupedExpressionFilters(token: Token, liquid: Liquid) {
  if (isGroupedExpressionToken(token)) {
    for (const t of token.initial.postfix) {
      resolveGroupedExpressionFilters(t, liquid)  // Recursive
    }
    token.resolvedFilters = token.filters.map(filterToken =>
      new Filter(filterToken, getFilter(liquid, filterToken.name), liquid)
    )
  }
  // ... handle Range/PropertyAccess recursively
}
```
Called from `Value` constructor, this builds `Filter` instances (which carry the `liquid` reference) at parse time for use at render time.

**4. Resolution calls in tag constructors** (regarding [this](https://github.com/harttle/liquidjs/pull/863#discussion_r3033452201)):

Tags that store raw `ValueToken` and evaluate them directly with `evalToken()` still need explicit `resolveGroupedExpressionFilters()` calls:
- `for.ts` - collection property
- `case.ts` - when values  
- `tablerow.ts` - collection property

Tags that wrap values with `new Value()` (like `if`, `unless`, `assign`, `echo`) don't need explicit calls since `Value` constructor handles resolution recursively via `initial.postfix`.

### Why This Works

- Maintains proper layering: tokens → render → templates
- Evaluation happens lazily at render time using generators (consistent with `Value.value()`)
- `Filter` instances created at parse time carry `liquid` reference for render-time filter resolution
- Minimal parse-time work: only build Filter instances, defer actual evaluation to render time

Attaching a draft PR with the proposed implementation - https://github.com/harttle/liquid
Does this approach align with your architectural vision?

jg-rp:
I was imagining that we could change the `Tokenizer` constructor to accept a `liquid` argument.

```ts
type TokenizerCtor = new (
  input: string,
  liquid: Liquid, // Get options directly from the Liquid object.
  file?: string,
  range?: [number, number]
) => Tokenizer;
```

And add `FilteredValueToken` to `ValueToken`.

```ts
type ValueToken =
    RangeToken
    | LiteralToken
    | QuotedToken
    | PropertyAccessToken
    | NumberToken
    | FilteredValueToken // This is new.
```

We'd then need to change `FilteredValueToken` to hold instances of `Filter` instead of `FilterToken`, and `Tokenizer.readFilter` would need to return `Filter` instead of `FilterToken`, which is now possible because it has access to `Tokenizer.liquid`.

(Notice that the existing definition of `FilteredValueToken` is almost identical to `GroupedExpressionToken`.)

To continue to support runtime tokenization in some filters, we'd also need to change `Context` to store a reference to the active `Liquid` object, allowing filters to create new tokenizers via `this.context.liquid`.

This approach moves responsibility for `Filter` construction from `Value` to `Tokenizer`, and avoids anything other than `Expression` knowing that sub expressions exist.

harttle:
Can we reuse `FilteredValueToken` and let `readGroupOrRange()` return `FilteredValueToken | RangeToken`? And `FilteredValueToken` can also be used to `new Value(filteredValueToken)` so it can be used in tag implementions with minor change.

I admit `ValueToken` and `Expression` are not well designed which hopefully can be addressed in v11. For now I don't think we can include filtered value into `ValueToken` as `ValueToken` is also used for `readProperty()` which should not accept a filtered value as the object of the property.

harttle:
@rosomri thank you for the effort. Comments to your proposals:

1. Token shouldn't depend on templates. Thus `GroupedExpressionToken` shouldn't have property of type `Filter`
2. `evalGroupedExpressionToken()` is good by itself. Need investigate can we reuse `FilteredValueToken` so this procedure can be reused from `Value#value()`.

3.,4. Same, let's see whether we can reuse.

jg-rp:
> Can we reuse FilteredValueToken and let readGroupOrRange() return FilteredValueToken | RangeToken? And FilteredValueToken can also be used to new Value(filteredValueToken) so it can be used in tag implementions with minor change.

Sounds good 👍 . We could even add an `ExpressionToken` (or similar) type alias:

```ts
export type ValueToken =
    RangeToken
    | LiteralToken
    | QuotedToken
    | PropertyAccessToken
    | NumberToken

export type ExpressionToken = ValueToken | FilteredValueToken;
```

There are quite a few places where `ValueToken` is used and would need to include `FilteredValueToken` (or similar), like `HashToken`, `RangeToken`, `FilterArg`. They should all probably allow arbitrary nesting of parenthesized expressions.

rosomri:
@harttle and @jg-rp, thank you for the detailed architectural feedback again! I've implemented the suggested improvements in a draft PR to demonstrate the approach: **#878**

## Changes in #878

**1. Reused `FilteredValueToken`** instead of `GroupedExpressionToken` (as suggested [here](https://github.com/harttle/liquidjs/pull/863#issuecomment-4236712239))
   - Simplified `readGroupOrRange()` to return `FilteredValueToken | RangeToken` 
   - Eliminated code duplication

**2. Fixed layering violation** (as noted [here](https://github.com/harttle/liquidjs/pull/863#discussion_r3033462685))
   - Removed `resolvedFilters?: Filter[]` from tokens (tokens no longer depend on template layer)
   - Added `liquid` reference to `Context` for runtime filter resolution (following @jg-rp's suggestion [here](https://github.com/harttle/liquidjs/pull/863#issuecomment-4236154245))

**3. Lazy filter resolution** at render time (as noted [here](https://github.com/harttle/liquidjs/pull/863#discussion_r3033458837))
   - Build `Filter` instances in `evalFilteredValueToken()` during rendering
   - Removed all `resolveGroupedExpressionFilters()` calls from parse time

## Results

✅ All 1537 tests pass  
✅ Proper architectural layering maintained: tokens → render → templates  
✅ Code simplified: -48 lines, removed duplicate class

## Question

The implementation adds `liquid` to `Context` for filter resolution. **Does this approach align with your architectural vision?** Or would you prefer passing `Liquid` to the `Tokenizer` constructor instead (which would require changes to filter runtime tokenization support)?

I'd appreciate your feedback on #878 to ensure the architectural direction is correct before finalizing this feature. Thanks!

jg-rp:
> Added liquid reference to Context for runtime filter resolution

👍 Render time filter resolution is a valid approach. This is what Shopify/liquid does.

I've drafted #879 as a proof of concept for maintaining parse-time filter resolution. I'm not yet convinced it is the best approach, just an option to consider.

harttle:
It’ll be good to avoid passing liquid instance to lower layers, if we can.
As previously FilteredValueToken also doesn’t check filter existence.

skynetigor:
Hi @harttle I did the changes, please give it another shot 🙏

In order to close all gaps, it seems we need major architectural change in LiquidJS to support this nice feature.
Would it be possible to merge it and take into account this use case during architectural shift in the next major version?

skynetigor:
@harttle did you have a chance to take another look?

src/parser/tokenizer.ts
```diff
@@ -307,10 +320,14 @@ export class Tokenizer {
...
-  readValue (): ValueToken | undefined {
+  readValue (): ValueToken | FilteredValueToken | undefined {
     this.skipBlank()
     const begin = this.p
-    const variable = this.readLiteral() || this.readQuoted() || this.readRange() || this.readNumber()
+    let variable: ValueToken | FilteredValueToken | undefined = this.readLiteral() || this.readQuoted()
+    if (!variable && this.peek() === '(') {
+      variable = this.readGroupOrRange()
+    }
+    variable = variable || this.readNumber()
```
harttle: why `readNumber` is special to move it after `readGroupOrRange`?

harttle: Had a quick look,it also returns `undefined` when no number is valid to read.

src/parser/tokenizer.ts
```diff
@@ -44,6 +51,11 @@ export class Tokenizer {
       return
     }
   }
+
+  * readGroupedExpressionTokens (lhs: Token): IterableIterator<Token> {
```
harttle: I don't understand why group expression tokens consists of a `lhs` and a list of other expression tokens.

The name `readExpressionTokensFromHere` is also wield.

src/parser/tokenizer.ts
```diff
@@ -385,18 +402,32 @@ export class Tokenizer {
...
+    if (this.peek() === '.' && this.peek(1) === '.') {
+      this.p += 2
+      const rhs = this.readValueOrThrow()
+      this.skipBlank()
+      this.assert(this.read() === ')', 'invalid range syntax')
+      return new RangeToken(this.input, begin, this.p, lhs, rhs, this.file)
+    }
+
+    if (this.groupedExpressions) {
+      const initial = new Expression(this.readGroupedExpressionTokens(lhs))
```
harttle: I see, we need reuse the lhs already parsed here.

Let's see if changing `Expression` argument type from `IterableIterator` to `Iterable` so array can be accepted. Then try something like this here:

```typescript
      const initial = new Expression([lhs, ...this.readExpressionTokens()])
```

If we can't change that type, we can try instead: 

```typescript
      const initial = new Expression([lhs, ...this.readExpressionTokens()].values())
```

harttle (changes requested):

harttle:
Thank you @skynetigor . The architecture is already good enough. I made some minor comments on some impl details.

skynetigor:
Hey guys!
@harttle I addressed all 3 last comments

harttle:
@all-contributors please add @skynetigor for code

PR #881, opened 2026-04-15
fix(filters): support Buffer input in base64_encode to prevent binary data corruption

talboren:
### Summary

When binary data (e.g. images, PDFs) is passed through the template context as a Node.js `Buffer`, the `base64_encode` filter corrupts the data. This is because:

1. `base64_encode` calls `stringify(value)` → which calls `String(value)` on the Buffer
2. `String(buffer)` invokes `Buffer.toString()` with the default `'utf-8'` encoding
3. UTF-8 decoding of arbitrary binary bytes is **lossy** — invalid byte sequences are replaced with U+FFFD (�)
4. The resulting base64 string encodes the corrupted text, not the original bytes

This is a follow-up to #828 which introduced the base64 filters.

### Use Case

This matters for template engines that process HTTP responses containing binary data. For example, downloading an image and base64-encoding it for an API call:

```liquid
{{ binary_image_data | base64_encode }}
```

Without this fix, the output is a base64 encoding of corrupted text rather than the original image bytes.

### Changes

- **`src/filters/base64.ts`** — Check `Buffer.isBuffer(value)` before `stringify()` to skip the lossy string conversion. Uses `value.byteLength` for memory limit accounting.
- **`src/filters/base64-impl.ts`** — Accept `string | Buffer` and call `buffer.toString('base64')` directly when given a Buffer.
- **`test/integration/filters/base64.spec.ts`** — 4 new tests covering Buffer encoding, invalid-UTF-8 byte preservation, empty Buffer, and UTF-8 text Buffer.

### Notes

- The browser implementation (`base64-impl-browser.ts`) is unchanged — browsers don't have `Buffer`, so `Buffer.isBuffer()` is always `false` in browser builds and the existing `btoa` path is used.
- All existing string-based tests continue to pass unchanged.
- The fix is fully backwards compatible.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/filters/base64-impl.ts
+++ b/src/filters/base64-impl.ts
@@ -1,5 +1,8 @@
-export function base64Encode (str: string): string {
-  return Buffer.from(str, 'utf8').toString('base64')
+export function base64Encode (value: string | Buffer): string {
+  if (Buffer.isBuffer(value)) {
+    return value.toString('base64')
+  }
+  return Buffer.from(value, 'utf8').toString('base64')
 }
 
 export function base64Decode (str: string): string {
--- a/src/filters/base64-impl.ts
+++ b/src/filters/base64-impl.ts
@@ -1,8 +1,5 @@
-export function base64Encode (value: string | Buffer): string {
-  if (Buffer.isBuffer(value)) {
-    return value.toString('base64')
-  }
-  return Buffer.from(value, 'utf8').toString('base64')
+export function base64Encode (str: string): string {
+  return Buffer.from(str, 'utf8').toString('base64')
 }
 
 export function base64Decode (str: string): string {
--- a/src/filters/base64.ts
+++ b/src/filters/base64.ts
@@ -8,7 +8,11 @@ import { FilterImpl } from '../template'
 import { stringify } from '../util'
 import { base64Encode, base64Decode } from './base64-impl'
 
-export function base64_encode (this: FilterImpl, value: string): string {
+export function base64_encode (this: FilterImpl, value: string | Buffer): string {
+  if (Buffer.isBuffer(value)) {
+    this.context.memoryLimit.use(value.byteLength)
+    return base64Encode(value)
+  }
   const str = stringify(value)
   this.context.memoryLimit.use(str.length)
   return base64Encode(str)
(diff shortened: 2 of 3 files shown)


talboren:
@harttle this was closed mistakenly. Would be happy if you can take a look

jg-rp:
I can confirm that Shopify/liquid's `base64_encode` filter does handle byte strings without clobbering the input's encoding, so this fix does improve Shopify/liquid compatibility. I do wonder if the fix can be generalized to work with browsers and `Uint8Array` though.

**Ruby**

```ruby
require "liquid"

some_bytes = String.new("\x89PNG\r\n\x1A\n\xFF\xFE", encoding: Encoding::ASCII_8BIT)

puts Liquid::Template
  .parse("{{ bytes | base64_encode }}")
  .render!({ "bytes" => some_bytes })

# iVBORw0KGgr//g==
```

**Aside**

I'm not a fan of anything that introduces byte data into the render context and usually prefer to remove all Base64 filters and wrap `Buffer` or `Uint8Array` in a `Blob` drop.

```js
import { Liquid, Drop } from "liquidjs";
import "core-js/proposals/array-buffer-base64.js";

const engine = new Liquid();

delete engine.filters.base64_encode;
delete engine.filters.base64_decode;

class Blob extends Drop {
  constructor(data) {
    super();
    this.data = data;
  }

  toString() {
    return "some bytes";
  }

  base64() {
    // Buffer inherits from Uint8Array.
    // Uint8Array.toBase64 is baseline since September 2025.
    // Works with browser and Node etc., with a pollyfill.
    // alphabet: "base64url" is available too.
    return this.data.toBase64({ alphabet: "base64", omitPadding: false });
  }
}

const someBytes = Buffer.from([
  0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a, 0xff, 0xfe,
]);

engine
  .parseAndRender("{{ bytes }}: {{ bytes.base64 }}", {
    bytes: new Blob(someBytes),
  })
  .then(console.log);

// some bytes: iVBORw0KGgr//g==
```

harttle:
Thanks guys. I think the MR is already good enough to be merged. @all-contributors please add @talboren for code.

PR #908, opened 2026-06-14
fix: json/inspect filters memory limit

harttle:


harttle:
I decided not to introduce non trivial fixes for memlimit, due to

1. We cannot fundamentally manage memory in Node.js app level
2. We can further simplify this with output limit which maybe more effective, see https://github.com/harttle/liquidjs/issues/910

The json/inspect check after string is generated only limits further additions, the current filter still run over limit.

Issue #866, opened 2026-04-01
Perf: Tokenizer rebuilds operator/literal tries on every instantiation


Kiryous:
## Problem

Every `new Tokenizer(...)` call rebuilds two tries from scratch via `createTrie()`:

```js
// tokenizer.ts constructor
this.opTrie = createTrie(operators);
this.literalTrie = createTrie(literalValues);
```

The `operators` and `literalValues` inputs are identical across all tokenizer instances for a given `Liquid` engine. In scenarios where many short template strings are parsed in a single pass (e.g. validating hundreds of `{{ variable }}` expressions in a document), this creates significant overhead from redundant trie construction.

## Profiling data

We're using LiquidJS (v10.25.0) inside [Kibana's workflow YAML editor](https://github.com/elastic/kibana) to validate Liquid template expressions on every keystroke. For a large document (~1,500 lines, ~360 template variables), CPU profiling shows:

- **`createTrie`: 173ms self-time** — the #10 hottest function by self-time
- It appears in **3 of the top 15 hot call stacks**, all through `Tokenizer → createTrie`
- Example stack: `parseTemplateString → parse → readTopLevelToken → readTagToken → TagToken → Tokenizer → createTrie`

We already cache `parse()` results (LRU by template string), which helps on cache hits. But on cache misses, `createTrie` is the dominant cost inside the tokenizer — more expensive than the actual tokenization work.

## Suggested fix

Cache the tries at the `Liquid` engine level (or `Parser` level) and pass them into `Tokenizer` instances, since the operator set and literal values don't change after engine construction:

```js
// In Parser or Liquid engine — build once
this.opTrie = createTrie(this.options.operators);
this.literalTrie = createTrie(literalValues);

// In Tokenizer constructor — reuse
constructor(input, opTrie, literalTrie, file, range) {
    this.opTrie = opTrie;
    this.literalTrie = literalTrie;
    // ...
}
```

This would be a non-breaking internal change. The `Tokenizer` constructor isn't part of the public API contract (users go through `Liquid.parse()`).

## Environment

- LiquidJS v10.25.0
- Browser (bundled via webpack)
- Hundreds of `parse()` calls per validation pass

PR #911, opened 2026-06-17
perf(parser): memoize createTrie to avoid rebuilding tries per Tokenizer

amedve:
## What

Memoize `createTrie` (`src/util/operator-trie.ts`) with a module-level `WeakMap` keyed on the input object, so the prefix-tries are built once per input and reused across `Tokenizer` instances.

## Why

The `Tokenizer` constructor calls `createTrie(operators)` and `createTrie(literalValues)` on **every** instantiation, and LiquidJS constructs a fresh `Tokenizer` per output/tag while parsing a template. On a typical template this rebuilds the identical tries dozens of times per parse, and the rebuild shows up as a significant share of parse CPU in profiling.

## Why this is safe

- The inputs are **stable references**: `literalValues` is a module-level constant, and `options.operators` is set once per `Liquid` instance.
- The trie is only ever **read** afterward (via `matchTrie`), never mutated, so caching by reference is behavior-preserving.
- `WeakMap` (not `Map`) is used so that short-lived, per-instance operator objects (e.g. the `{}` literal passed in `Hash`), and their tries can be garbage collected. No unbounded growth.
- A comment documents that the returned trie must be treated as read-only, since `createTrie` is re-exported from `src/index.ts` (public API).

## Benchmarks

Built with `npm run build:cjs` and compared against `liquidjs@10.27.0` from unpkg, on `benchmark/templates/todolist.liquid` (arm64, Node, 5s windows):

| Path | latest 10.27.0 | local (this PR) | change |
|---|---|---|---|
| **parse** (`engine.parse`) | 18,807 parse/s (53.2 µs) | 37,194 parse/s (26.9 µs) | **+97.8% (~2x)** |
| **render** (`npm run perf:diff`) | 24,996 ops/s | 25,008 ops/s | +0.047% (noise) |

The official `perf:diff` harness parses the template once outside the timed loop and only times `render`, so it (correctly) shows no change for a parse-path optimization, i.e. no render regression. The render path is unaffected; the win is in parse/`engine.parse`, which the table above isolates.

## Tests

- Added a `#createTrie()` block in `src/parser/tokenizer.spec.ts` asserting that the same input returns the same trie reference and distinct inputs return distinct tries.

### DIFF
--- a/src/parser/tokenizer.spec.ts
+++ b/src/parser/tokenizer.spec.ts
@@ -522,6 +522,14 @@ describe('Tokenizer', function () {
       expect(new Tokenizer('contains b').matchTrie(opTrie)).toBe(8)
     })
   })
+  describe('#createTrie()', function () {
+    it('should return the same trie for the same input', () => {
+      expect(createTrie(defaultOperators)).toBe(createTrie(defaultOperators))
+    })
+    it('should return distinct tries for distinct inputs', () => {
+      expect(createTrie({ foo: 1 })).not.toBe(createTrie({ foo: 1 }))
+    })
+  })
   describe('#readLiquidTagTokens', () => {
     it('should read newline terminated tokens', () => {
       const tokenizer = new Tokenizer('echo \'hello\'')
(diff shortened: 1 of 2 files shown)


harttle:
Relates to #866, the change seems acceptable. Merging and planning for thorough fix on v11.

harttle:
@all-contributors please add @amedve for code

PR #912, opened 2026-06-19
fix(security): reject render/include partial recursion

harttle:
Detect cyclic partial rendering via a shared partialStack register (same pattern as CVE-2026-41311 block tag fix). Self-referential or circular {% render %} and {% include %} now throw immediately instead of hanging or OOM.


harttle:
closing as recursion need to be allowed and we have `renderLimit` which should be used instead.

PR #928, opened 2026-07-08
fix(filters): return empty for out-of-range slice begin or negative length

spokodev:
Ruby/Shopify `slice` returns nil (rendered as `""`/`[]`) when the begin offset is outside the negative range or the length is negative. liquidjs forwarded the adjusted indices straight to `Array/String.prototype.slice`, whose own negative-index handling produced non-empty, incorrect output:

```
{{ "hello" | slice: -10, 2 }}  → "he"    (expected "")
{{ "Liquid" | slice: 1, -2 }}  → "iqui"  (expected "")
```

Guard the adjusted begin and the length before slicing.


harttle:
@all-contributors please add @spokodev for code

PR #939, opened 2026-07-20
fix(url_decode): keep %2B as a literal plus when decoding

sarathfrancis90:
I noticed that `url_decode` loses a literal `+`: `{{ "1%2B1" | url_decode }}` gives `1 1` instead of `1+1`, and anything with a `+` no longer survives a `url_encode | url_decode` round-trip (`a+b c` comes back as `a b c`).

The cause is ordering: it ran `decodeURIComponent` first and only then replaced every `+` with a space, so a `%2B` decoded to `+` and was immediately turned into a space. I moved the `+`-to-space replacement ahead of `decodeURIComponent`, so encoded pluses survive. This matches Ruby's `CGI.unescape` that Shopify uses.

Tested with two new cases in `test/integration/filters/url.spec.ts` (the `%2B` decode and the round-trip), plus the full suite and lint.

### DIFF
--- a/src/filters/url.ts
+++ b/src/filters/url.ts
@@ -1,6 +1,6 @@
 import { stringify } from '../util/underscore'
 
-export const url_decode = (x: string) => decodeURIComponent(stringify(x)).replace(/\+/g, ' ')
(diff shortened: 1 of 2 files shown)


harttle:
@all-contributors please add @sarathfrancis90 for code.

PR #943, opened 2026-07-28
feat(filters): add squish filter

MildlyMeticulous:
`squish` is a standard Shopify/liquid string filter and LiquidJS doesn't have it, so `{{ "Hello   World!" | squish }}` currently renders unchanged rather than `Hello World!`.

```js
engine.parseAndRenderSync('{{ "  Hello   World!  " | squish }}')
// before: "  Hello   World!  "
// after:  "Hello World!"
```

Strips leading and trailing whitespace and collapses each internal run to a single space, matching Shopify/liquid 5.12.0. Added alongside `strip_newlines` in `src/filters/string.ts`, with docs and a sidebar entry.

Verified against all 17 `squish` cases in [golden-liquid](https://github.com/jg-rp/golden-liquid), which tests against Shopify/liquid as the reference implementation. All pass. 1588 tests pass here; the new tests fail on revert.


src/filters/string.ts
```diff
@@ -128,6 +128,12 @@ export function strip_newlines (this: FilterImpl, v: string) {
   return str.replace(/\r?\n/gm, '')
 }
 
+export function squish (this: FilterImpl, v: string) {
+  const str = stringify(v)
+  this.context.memoryLimit.use(str.length)
+  return str.replace(/\s+/g, ' ').trim()
```
harttle, line 134: Not sure if the spaces trimed by `.trim()` is same with those `\s` matches, and whether the order of replace and trim matters (it's different with shopify).

Anyway this is good enough to merge. Will handle it when discrepancies arise.

PR #946, opened 2026-08-09
feat: add unregisterFilter method

YacovGold:
## Summary

- Add `Liquid#unregisterFilter(name)`
- Support removing custom and built-in filters
- Allow filters to be registered again after removal
- Document restoring built-in filters through the exported `filters` object
- Document behavior with `strictFilters`

## Motivation

This provides a supported API for removing filters without directly modifying the filter registry.

When `strictFilters` is disabled, an unregistered filter is skipped. When enabled, using it throws an undefined-filter error. Unregistering an unknown
filter is a no-op.

Built-in filters can be restored using the existing `registerFilter()` API:

```javascript
import { filters } from 'liquidjs'

engine.registerFilter('plus', filters.plus)


harttle:
@all-contributors please add @YacovGold for code

PR #954, opened 2026-09-15
fix(truncatewords): ignore leading whitespace when counting words

youdie006:
`truncatewords` does `str.split(/\s+/)` (`src/filters/string.ts:183`). When the input starts with whitespace that yields a leading `''` element, which gets counted as a word and joined back into the output:

```
{{ "  Ground control to Major Tom." | truncatewords: 3 }}
liquidjs -> " Ground control..."
Ruby     -> "Ground control to..."
```

So the result is one word short and carries a stray leading space. This shows up on CMS/HTML-derived text and `{% capture %}` output, which routinely start with whitespace.

`trimStart` rather than `trim`, because Ruby's awk-mode `split(" ", n+1)` discards leading whitespace but leaves trailing whitespace alone - `{{ "1 2 3 " | truncatewords: 5 }}` returns `"1 2 3 "` in Ruby, and `trim` would drop that space. Both test rows below cover that.

`number_of_words` is unaffected: it already calls `.trim()` at `string.ts:199` before splitting.

<details>
<summary>Verification</summary>

Oracle measured against the `liquid` gem 5.13.0 on `ruby:3.3`:

| input | words | Ruby | liquidjs before |
|---|---|---|---|
| `"  Ground control to Major Tom."` | 3 | `"Ground control to..."` | `" Ground control..."` |
| `"\t one two three four"` | 2 | `"one two..."` | `" one..."` |
| `"1 2 3 "` | 5 | `"1 2 3 "` | `"1 2 3 "` |

`TZ=Asia/Shanghai npm test` (the command `test.yml` runs): **89 suites, 1644 tests, all passing**. `npm run lint` exit 0.

Reverting the one-line change fails the first new test (`Expected: "Ground control to..." / Received: " Ground control..."`); using `trim()` instead of `trimStart()` fails the second (`Expected: "1 2 3 " / Received: "1 2 3"`).

I left the existing `{{ "1 2 3" | truncatewords: 3 }}` -> `'1 2 3...'` row alone; Ruby returns `"1 2 3"` there, but that ellipsis policy is pinned by your own test and is a separate question from this one.

</details>

I used Claude Code while working on this; the change and the tests are mine and I have been through them.


harttle:
@all-contributors please add @youdie006 for code

PR #955, opened 2026-09-19
fix: trim custom characters by Unicode code point

jakezwang:
Custom characters are already collected as Unicode code points, but `strip`, `lstrip`, and `rstrip` compare them against individual UTF-16 code units. For example, `{{ "𠮷text𠮷" | strip: "𠮷" }}` returns the unchanged input instead of `text`.

Scan complete code points at each boundary and retain string offsets for slicing. This avoids copying the full input and preserves lone surrogates. Add async/sync rendering regressions for mixed characters, empty inputs, and distinct characters sharing a surrogate.

Validation: build and lint pass; all 1,666 tests pass on Node 16, 20, and 24. Node 22 coverage and demo checks pass.


harttle:
@all-contributors please add @jakezwang  for code.

